PR #1, opened 2015-06-19
Add configuration and authentication token functionality.

eschlange:
* Added all necessary controllers, resources, services, and constants for token creation.
* Removed unnecessary files copied over from other Cordova projects.
* Updated configurations and index page to reference all of the new files.

[#88080168]


config.xml
```diff
@@ -1,12 +1,13 @@
 <?xml version='1.0' encoding='utf-8'?>
```
ericcf: I would suggest adding this to `.gitignore` because it will get regenerated each time you build the app for a different backend: local, staging, production.

package.json
```diff
@@ -38,18 +39,19 @@
...
     "appium:start": "./node_modules/.bin/appium 2>/dev/null 1>&2 &",
     "appium:stop": "ps -A | grep appium | grep -v grep | egrep -o \"^\\d+\" | xargs kill",
-    "browser:add_platform": "./node_modules/.bin/cordova platform remove browser && ./node_modules/.bin/cordova platform add browser",
+    "browser:add_platform": "./node_modules/.bin/cordova platform remove browser && ./node_modules/.bin/cordova platform add browser && npm run add_plugins",
+    "add_plugins": "./node_modules/.bin/cordova plugin remove org.apache.cordova.device; ./node_modules/.bin/cordova plugin add org.apache.cordova.device; ./node_modules/.bin/cordova plugin remove org.apache.cordova.network-information; ./node_modules/.bin/cordova plugin add org.apache.cordova.network-information; ./node_modules/.bin/cordova plugin remove de.appplant.cordova.plugin.local-notification; ./node_modules/.bin/cordova plugin add de.appplant.cordova.plugin.local-notification; ./node_modules/.bin/cordova plugin remove com.phonegap.plugins.PushPlugin; ./node_modules/.bin/cordova plugin add https://github.com/Telerik-Verified-Plugins/PushNotification.git",
```
ericcf: Do you think you'll need push notifications for this project? If not, you may as well remove the PushNotification plugin.

ericcf:
Looks good. Just needs unit tests and an Appium test. There should be plenty of examples in ed nav and medlink to make progress with those. Up to you if you want to add eslint on this PR or wait until that chore comes up. Thanks!

config.xml
eschlange: I must have checked this in before updating the ignore file. I'll remove the file from the project.

package.json
eschlange: There should be two push notifications per day. One prompts for the morning EMA input, the other is for the evening EMA input.

eschlange:
@ericcf Please take a look at my latest revisions and let me know if I am missing anything essential for the test framework. I copied over and modified an array of unit tests for files similar to those found in the other projects and was able to get them all running locally.

package.json
ericcf: We should probably discuss this requirement. I was under the impression that this app needed to work offline as well.

ericcf:
Tests run locally for me as well. Don't forget the Travis config so they run on CI.

package.json
```diff
@@ -38,18 +39,19 @@
     "url": "https://github.com/cbitstech/sis_app"
   },
   "scripts": {
+    "set_endpoint": "echo \"{\\\"environment\\\":\\\"$npm_config_env\\\",\\\"server\\\":\\\"$npm_config_server\\\"}\" > src/config.json && sed s%ACCESS_ORIGIN%$npm_config_server%g config > config.xml",
     "android:add_platform": "./node_modules/.bin/cordova platform remove android && ./node_modules/.bin/cordova platform add android@4.0.0",
     "android:build": "npm run build:prepare && ./node_modules/.bin/cordova build android",
     "android:simulator": "npm run android:build && ./node_modules/.bin/cordova run android",
```
michaelwehrley: I was playing with this at home this weekend, and given what @ericcf pointed out in the past, you actually only need to run `npm run build:prepare && ./node_modules/.bin/cordova run android` because `run android` also runs the command `build android`.  Not a huge deal, but it may speed up the emulator.

eschlange:
@ericcf All updates are complete, would you please merge this when you get a chance?

PR #3, opened 2015-06-25
Add cessation date configuration and risky time CRUD.

eschlange:
* Added template for home screen.
* Added template for config screen.
* Updated selector for cessation time to use angular.
* Update controller logic for cessation time selection.
* Added risky times controller and route.
* Added page for selection of risky times.
* Added spec coverage for the two new controllers.

[#95521946]


src/partials/home.html
```diff
@@ -1,11 +1,85 @@
...
+<div id="app-container" class="ng-scope">
+    <div id="setup-menu" class="ng-scope">
+        <a ng-href="#/set_up" title="Set Up" href="#/set_up">
+            <i class="glyphicon glyphicon-cog glyphicon-lg"></i>
+        </a>
+    </div>
+
+    <div class="container ng-scope">
+        <header id="smokingStatus" class="text-center before">
+            <span class="days ng-binding">8</span> <span class="suffix ng-binding">days until quit day!</span>
```
ericcf: Did you mean to add the `ng-binding` class here? It looks like Angular will add/remove it automatically.

eschlange: It got copied in from Mark's prototype. I'll remove it.

src/partials/risky_times.html
```diff
@@ -0,0 +1,68 @@
+<div id="app-container" class="ng-scope">
+    <div class="container ng-scope">
+        <h2>Your Risky Times</h2>
+        <h4>These are times you have indicated you might have particular trouble abstaining from smoking:</h4>
+
+        <div class="well">
+            <span ng-if="!risky.riskyTimes || risky.riskyTimes == []">Add risky times below.</span>
```
ericcf: I don't think this is going to work the way you expect. In JS, `[] == []` will return false. It may be simpler to initialize riskyTimes to `[]` and then just check for `riskyTimes.length === 0`.

src/partials/risky_times.html
```diff
@@ -0,0 +1,68 @@
...
+            <div class="btn-group" role="group">
+                <button id="weekday_button_su"
+                        type="button"
+                        class="btn btn-default weekday_button"
+                        ng-click="risky.selectWeekday(risky.SUNDAY_KEY)"
+                        ng-class="{'active': risky.activeButton == risky.SUNDAY_KEY}">
+                    Su
+                </button>
+            </div>
+            <div class="btn-group" role="group">
```
ericcf: Looks like a prime candidate for `ngRepeat`. That will make it much easier to read and/or make changes I think.

src/partials/risky_times.html
```diff
@@ -0,0 +1,68 @@
...
+        <textarea ng-model="risky.riskyTime.description" class="form-control" name="reason" placeholder="e.g. people are smoking during break"></textarea>
+
+        <br>
+
+        <a id="save_button" class="btn btn-info" ng-click="risky.saveRiskyTime()">Save</a>
+    </div>
+</div>
+
+<script type="text/javascript">
+  $("#risky_time_time").mobiscroll().time();
```
ericcf: Since you're using this in several places, I'd consider using a directive if the logic gets any more complex or you have to make changes. May not be worth the effort yet though.

src/partials/home.html
```diff
@@ -1,11 +1,85 @@
...
+    </div>
+
+    <div class="row top-buffer20">
+        <div class="col-md-6 col-sm-6 col-lg-6 col-xs-6 text-center"> Today's Happiness Exercise is: </div>
+        <div class="col-md-6 col-sm-6 col-lg-6 col-xs-6 text-center"> Quitting Smoking session: </div>
+    </div>
+
+    <div class="row">
+        <div class="col-md-6 col-sm-6 col-lg-6 col-xs-6 text-center">
+            <a ng-href="#/good_things" class="wide btn btn-default btn-lg" href="#/good_things">
```
michaelwehrley: I don't believe you shouldn't need both the `ng-href` and `href`

eschlange:
Cool, thanks guys. Mind if I merge this in?

PR #4, opened 2015-06-26
Add cessation day information to home page.

eschlange:
* Updated homecontroller to include new date calculation functions.
* Added specs for the new homecontroller functions.
* Updated home template.

[#95521898]


src/js/controllers/HomeController.js
```diff
@@ -1,10 +1,25 @@
 (function() {
   'use strict';
 
-  function HomeController($timeout) {
+  function HomeController() {
+      this.CESSATION_DATE_KEY = 'cessation_date';
+
+      this.cessationDate = JSON.parse(localStorage.getItem(this.CESSATION_DATE_KEY));
```
ericcf: If you use `angular.fromJson` it's one less global dependency: https://docs.angularjs.org/api/ng/function/angular.fromJson

src/partials/home.html
```diff
@@ -1,13 +1,31 @@
...
-            <span class="days">8</span> <span class="suffix ">days until quit day!</span>
+            <div ng-if="home.cessationDayOffset() < 0">
+                <span class="days">{{home.cessationDayOffsetAbs()}}</span>
+                <span class="suffix">
+                    <ng-pluralize count="home.cessationDayOffsetAbs()"
+                                  when="{'one': ' day until quit day!','other': ' days until quit day!'}">
+                    </ng-pluralize>
+                </span>
+            </div>
+            <div ng-if="home.cessationDayOffset() == 0">
```
ericcf: I recommend never using double equals (or `!=`) in JavaScript, even in trivial comparisons. If you always use `===` and `!==` you'll never accidentally type cast an argument.

PR #5, opened 2015-06-26
A cessation session display to home page.

eschlange:
* Updated cessation session display.
* Updated controller for home.
* Added specs for new unit tests.

[#95521876]


src/js/controllers/HomeController.js
```diff
@@ -4,6 +4,11 @@
...
       this.CESSATION_DATE_KEY = 'cessation_date';
 
+      // Cessation session states
+      this.CESSATION_SESSION_PRE = 'pre_cessation';
+      this.CESSATION_SESSION_DURING = 'cessation';
+      this.CESSATION_SESSION_POST = 'post_cessation';
+
       this.cessationDate = JSON.parse(localStorage.getItem(this.CESSATION_DATE_KEY));
 
       this.cessationDayOffset = function() {
```
michaelwehrley: I just saw this (https://github.com/cbitstech/sis_app/pull/5/files#diff-d4c71c7acb5302444bc48119d4d59f23L12), but you may consider using 
`moment().startOf('day');` instead of `moment((new Date()).setHours(0,0,0,0));` Not sure how timezones work with what you are doing, but it seems to be simpler.

test/unit/controllers/HomeControllerSpec.js
```diff
@@ -37,4 +37,18 @@ describe('HomeController', function() {
         });
     });
 
+    describe('#getCurrentSession', function () {
+        it('returns pre-cessation session if current day is before cessation date', function () {
+            controller.cessationDate = moment(new Date()).subtract(7, 'days');
```
michaelwehrley: Do you need `new Date()`?  That is, I think you can write it simply as `moment().subtract(7, 'days');`.  Not a huge deal, but it makes things a bit simpler and easier to understand.

eschlange:
@michaelwehrley Thanks for the comments, Mike! I updated the controller and spec code to reflect your recommended changes. That is a useful constructor!

michaelwehrley:
:eyeglasses: good! :+1:

PR #6, opened 2015-06-29
Add cessation reason configuration.

eschlange:
* Added new controller.
* Added specs for controller.
* Added page.
* Added configuration for all new files.

[#95521996]


src/js/controllers/CessationReasonsController.js
```diff
@@ -0,0 +1,36 @@
+(function() {
+    'use strict';
+
+    function CessationReasonsController() {
+        this.CESSATION_REASONS_KEY = 'cessation_reasons';
+
+        this.cessationReason = {};
+        this.cessationReasons = localStorage.getItem(this.CESSATION_REASONS_KEY) ? JSON.parse(localStorage.getItem(this.CESSATION_REASONS_KEY)) : [];
```
ericcf: Is there a reason for eagerly loading `cessationReasons` from local storage as opposed to lazily?

eschlange: I'm displaying the reasons wherever this controller is used, but I can see your point. Should I simply create an accessor which does this work?

ericcf: That would be my approach because `cessationReasons` is mutable.

eschlange: OK, making the update now.

eschlange:
@ericcf I made the controller more functional and changed it to lazily load cessation reasons.

ericcf:
:ice_cream:

PR #8, opened 2015-06-29
Add ability to create risky smoking time strategies.

eschlange:
* Add new controller.
* Add new view.
* Add specs for new controller.
* Updated config to include new files.
* Added new route.

[#95521992]


src/js/controllers/RiskyTimeStrategiesController.js
```diff
@@ -0,0 +1,37 @@
+(function() {
+    'use strict';
+
+    function RiskyTimeStrategiesController() {
+        this.RISKY_TIME_STRATEGIES_KEY = 'risky_time_strategies';
```
michaelwehrley: why did set it to `this` instead of using `var RISKY_TIME_STRATEGIES_KEY = 'risky_time_strategies';`

eschlange: I added this to the parent class in order to allow other classes a reference to the stored values. This will be abstracted out of this class completely once I refactor to use the caches.

michaelwehrley: Or create another constants file?

michaelwehrley:
:eyeglasses: good to me! :ship: :hear_no_evil:

PR #9, opened 2015-07-01
Refactor storage methods to allow for config passing among controllers.

eschlange:
* Updated each controller to include caching logic.
* Added the configuration service.
* Updated the unit test runner to include config.
* Updated specs to include new logic.
* Updated views to use new config methods.

[#88080168]


src/partials/home.html
```diff
@@ -22,7 +22,7 @@
                 <span class="days">{{home.cessationDayOffset()}}</span>
                 <span class="suffix">
                     <ng-pluralize count="home.cessationDayOffset()"
-                                  when="{'one': ' day since you quit!','other': ' days since you quit!'}">
+                                  when="{'one': ' day until quit day!','other': ' days until quit day!'}">
```
michaelwehrley: I would not want an explanation point there :exclamation: lol

src/js/controllers/SetUpController.js
```diff
@@ -1,21 +1,24 @@
...
+          return configurationService.getCessationDate();
         };
 
         this.setCessationDate = function() {
-            localStorage.setItem(this.CESSATION_DATE_KEY, JSON.stringify(this.cessationDate));
+          configurationService.saveCessationDate(this.cessationDate);
         };
+
+        this.configurationComplete = function() {
+          debugger
```
michaelwehrley: note.

src/partials/home.html
cbrenner04: I think you mean exclamation mark?

eschlange: Just following design artifacts~

src/js/controllers/SetUpController.js
eschlange: removing

michaelwehrley:
looks like good refactoring here! :+1:

src/partials/home.html
michaelwehrley: Dude, you are so right.  Whoops!

eschlange:
@michaelwehrley Thanks, Mike! Should I merge this in?

PR #10, opened 2015-07-02
Add routing logic to force navigation to setup page.

eschlange:
[#88080168]


src/js/Application.js
```diff
@@ -43,11 +43,15 @@
...
                       authenticationTokenCache, Routes, synchronizer,
-                      deviceCache, device) {
+                      deviceCache, device, configurationService) {
         $q.all(settings.fetch()).then(function() {
 
             // if (authenticationTokenCache.first()) {
+            if(configurationService.configurationComplete())
               $location.url(Routes.HOME);
+            else {
+              $location.url(Routes.SET_UP);
```
michaelwehrley: What does `$location.url()` do?  When I have changed the path, I usually use `$location.path()`.  From the documentation (https://docs.angularjs.org/guide/$location) it says:
```javascript
// change the path
$location.path('/newValue')
```
Thanks!

eschlange: @michaelwehrley They do the same thing, with the exception of: `$location.path() setter will add the forward slash if it is missing`

I use ".url" because I always want to specify the exact path for consistency and assurance of the destination path.

michaelwehrley:
:+1:

PR #11, opened 2015-07-06
Add daily exercises.

eschlange:
* Added specified exercise descriptions and content.
* Added cache for exercise data.
* Added exercise service.
* Added view for exercises.
* Added controller for exercises.
* Added specs for exercise cache and controller methods.
* Updated home view to use new exercise data.
* Updated routes and configuration files for new logic.

[#95522472]


src/js/services/Exercise.js
```diff
@@ -0,0 +1,63 @@
...
+    };
+
+    this.saveNewExercise = function (exercise) {
+      exerciseCache.persist(exercise);
+    };
+
+    this.updateExercise = function (exercise) {
+      debugger
+      var exerciseData = exerciseCache.fetchAllRaw();
+      for (var i = 0; exerciseData && i < exerciseData.length; i++) {
```
michaelwehrley: You have debugger on line 42, and could you use https://docs.angularjs.org/api/ng/function/angular.forEach here?  If you chose the incrementing over the each method, why?  Thanks.
```javascript
angular.forEach(exerciseData, function(value, key) {
```

test/unit/controllers/ExercisesControllerSpec.js
```diff
@@ -0,0 +1,35 @@
...
+    it('should save new exercise answers', function() {
+      controller.todaysExercise = {};
+      controller.todaysExercise.type  = "Three Good Things";
+      controller.todaysExercise.date = moment();
+      controller.todaysExercise.description = "This is a description";
+      controller.todaysExercise.answers = [];
+      controller.todaysExercise.answers.push("some answer");
+      controller.todaysExercise.answers.push("another answer");
+      controller.saveExercise();
+      expect(controller.todaysExercise.answers.length).to.equal(2);
```
michaelwehrley: Because answers is an array, by pushing all this into the array, if you removed line 23 would it still pass?

michaelwehrley:
:eyes: great to me!

src/js/services/Exercise.js
eschlange: I could, but then I'd have to write a wrapping conditional to check for an undefined exerciseData, angular expects the object passed in to not be undefined. The for loop takes care of both.

ericcf:
@michaelwehrley do you think unit tests are needed for the `Excercise` service?

src/js/services/Exercise.js
ericcf: Maybe we should update `fetchAllRaw` to return an empty array instead. You could also use

```javascript
angular.forEach((exerciseData || []), function(value, key) {
```

eschlange:
@ericcf @michaelwehrley I forgot to include my exercises for the 'Exercise' service. I'll commit and push those shortly.

eschlange:
@michaelwehrley Please take a look at my changes. I made the updates that you recommended, added exercise service test cases, and refactored my existing ones.

michaelwehrley:
:+1:

PR #14, opened 2015-07-08
Add cigarette log functionality

eschlange:
* Added controller, service and cache to handle cigarette logs.
* Added specs to test new classes.
* Added view for logging cigarettes.
* Updated configuration files to include new classes.

[#95522106]


config
```diff
@@ -1,8 +1,8 @@
 <?xml version='1.0' encoding='utf-8'?>
-<widget id="edu.northwestern.cbits.sis" version="0.2.0" xmlns="http://www.w3.org/ns/widgets" xmlns:cdv="http://cordova.apache.org/ns/1.0">
+<widget id="edu.northwestern.cbits.sis" version="0.3.0" xmlns="http://www.w3.org/ns/widgets" xmlns:cdv="http://cordova.apache.org/ns/1.0">
     <name>SiS</name>
     <description>
-        An application to help study medication adherence.
+        An application to aid in the cessation of smoking cigarettes.
```
ericcf: :+1:

src/partials/cigarette_log.html
```diff
@@ -0,0 +1,20 @@
+<div id="app-container">
+    <div class="container">
+        <h2>Cigarette Log</h2>
+        <button ng-click="smoked = true && cigarette.saveCigaretteLog()" ng-show="smoked == null" class="btn btn-lg btn-info" tabindex="0" aria-hidden="false">
```
ericcf: How/where does the variable `smoked` get set to `null`?

src/partials/cigarette_log.html
```diff
@@ -0,0 +1,20 @@
...
+            I'm smoking now...
+        </button>
+        <br>
+        <button ng-click="smoked = false && cigarette.saveCigaretteLog()" ng-show="smoked == null" class="btn btn-lg btn-info" tabindex="0" aria-hidden="false">
+            I forgot to enter a cigarette...
+        </button>
+        <div ng-show="smoked == true" aria-hidden="true" class="ng-hide">
```
ericcf: Can this be simplified to `ng-show="smoked"`? I haven't tried, but I think it should.

src/partials/cigarette_log.html
eschlange: This logic is more of a placeholder for what is to come, I copied this logic in from Mark's prototype. Once we add the EMA questions into this flow, I'll likely remove this hide and show logic completely.

ericcf:
Sounds good. Thanks! :pushpin:

PR #20, opened 2015-07-14
Update logic to use newly defined session dates.

eschlange:
* Session 3 now begins 7 days after cessation.
* Updated controller and view logic with the new day calculations.

[#95522324]


src/partials/home.html
```diff
@@ -101,15 +101,15 @@
 
         <div class="col-md-6 col-sm-6 col-lg-6 col-xs-6 text-center">
             <div class="label label-lg label-info"
-                 ng-if="home.getCurrentSession() === home.CESSATION_SESSION_PRE">
-                session 3 in {{home.cessationDayOffsetAbs() + 1}} days
+                 ng-if="home.cessationDayOffset() > -5">
```
michaelwehrley: You mentioned you will update these to constants in the future.  Therefore, all else looks good!

PR #21, opened 2015-07-14
Create temptation logging and update smoking statuses.

eschlange:
* converted existing temptation log features to smoking statuses.
* created new controller, services and view specific to temptation logging.
* created specs for the new services and controller.

[#95522168]


src/js/services/CigaretteLog.js
```diff
@@ -3,11 +3,8 @@
 
   function CigaretteLog(cigaretteLogCache) {
 
-    this.saveCigaretteLog = function (description) {
-      var cigaretteLog = {
-        'date': JSON.stringify(moment()),
-        'description': description
-      };
+    this.saveCigaretteLog = function (cigaretteLog) {
+      cigaretteLog.date = JSON.stringify(moment());
```
ericcf: Alternatively, you could call `moment().toISOString()`.

eschlange: I noticed you using that method in medlink_2. I'll probably use that in the future for date marshalling.

ericcf:
:rotating_light:

PR #22, opened 2015-07-20
Add session logic, refactor existing pages to use session service.

eschlange:
* Added necessary controllers, services, and cache.
* Updated eslint config for raw data files.
* Added and updated views for the new functionality.
* Added specs for the new service, controller, and cache.
* Updated home page logic to use new session service.
* Updated configurations to include new files.
* Updated views to utilize new controller logic.

[#95522266]


eschlange:
*NOTE* to reviewer: you can ignore the big file, "src/js/constants/SessionContent.js". This file is being imported from REDCap.

src/js/constants/SessionContent.js
```diff
@@ -0,0 +1,1304 @@
+var sessionContent = {};
```
ericcf: Is there a reason this needs to be a global?

src/js/controllers/SessionsController.js
```diff
@@ -0,0 +1,79 @@
...
+    this.sessionStarted = false;
+
+    this.sessionKey = undefined;
+    this.currentAnswer = undefined;
+    this.currentQuestion = undefined;
+    this.sessionData = {};
+
+    this.sessionContent = undefined;
+    this.sessionName = '';
+    this.sessionDescription = 'Click the button to begin...';
```
ericcf: Why did you decide to store this content on a controller property?

src/js/services/Session.js
```diff
@@ -0,0 +1,95 @@
...
+      sessionsCache.persist(sessionData);
+    };
+
+    // Returns the cessation session state relative to an
+    // offset in days for the cessation day.
+    this.cessationSession = function (offset) {
+      var session;
+      if (offset > 0) {
+        session = this.CESSATION_SESSION_PRE;
+      } else if (offset <= 0 && offset > -6) {
```
ericcf: What do you think about creating a named constant for the "magic" value `-6`? I think that would help readability.

src/js/services/SessionQuestion.js
```diff
@@ -0,0 +1,153 @@
+(function() {
+  'use strict';
+
+  function SessionQuestion() {
+    this.redCapJSONContent = undefined;
```
ericcf: Why did you choose this name for the property?

src/js/services/SessionQuestion.js
```diff
@@ -0,0 +1,153 @@
...
+              self.currentIndex === index;
+    };
+    /*eslint-enable */
+
+
+    this.next = function(){
+      var self = this;
+      var selected = [];
+
+      $('input[name="' + self.content[this.currentIndex].variable + '"]:checked').each(function() {
```
ericcf: Is it possible to use `angular.element` so there's not a direct jQuery dependency?

src/js/services/SessionQuestion.js
```diff
@@ -0,0 +1,153 @@
+(function() {
```
ericcf: Are there unit tests for this service? It looks pretty complex.

src/js/constants/SessionContent.js
eschlange: This was the file given to me by Mark. I'll check with him to see if we can modify the output and formatting of the SessionContent.js file.

src/js/controllers/SessionsController.js
eschlange: This is more of Mark's code. I'll refactor this into the view, since I don't believe that it needs to be dynamic.

src/js/services/Session.js
eschlange: The constant already existed in the home service, I'll abstract it into the session service and delegate the home service references to the session service.

src/js/services/SessionQuestion.js
eschlange: Mark chose this name, I'll refactor it.

src/js/services/SessionQuestion.js
eschlange: I couldn't get it working so I did some searching. I found this article which references the Angular documentation that states that this is not possible without jumping through some hoops: http://stackoverflow.com/a/17230735/512463. Most of the posts recommend including the full jQuery lib if using selectors.

eschlange:
@ericcf I made your recommended changes. I copied in Mark's tests for his parsing service, removed the "magic numbers" and refactored the questionable variable names.

ericcf:
I'm getting 34/80 lines covered for the SessionQuestion service. Is that what you have? Do you think that's sufficient and/or is some of the code unnecessary?

eschlange:
@ericcf I'm not sure if we'll benefit much from putting more testing into this parser. The session content is not yet finalized.

ericcf:
Ok, I'll just write up a separate story for it if you prefer.

eschlange:
@ericcf I can write it up. Thanks!

PR #23, opened 2015-08-05
Evening and morning EMA notifications.

eschlange:
* Added morning notifications for EMA input.
* Added evening notifications for EMA input.
* Added EMA service and controllers.
* Added spec coverage to all new files.
* Small update to session content and display.
* Added common EMA questions used by both morning and evening notification views.

[#95522128]


src/js/Application.js
```diff
@@ -65,25 +65,59 @@
...
+          controllerAs: 'ema'
         });
     },
 
     run: function run($rootScope, $location, $q, $window, settings,
                       authenticationTokenCache, Routes, synchronizer,
-                      deviceCache, device, configurationService) {
+                      deviceCache, device, configurationService,
+                      emaService) {
         $q.all(settings.fetch()).then(function() {
```
ericcf: Why does `settings.fetch()` need to be wrapped with `.all`?

src/js/Application.js
```diff
@@ -65,25 +65,59 @@
...
             }
-            // } else {
-            //   $location.url(Routes.CONFIG);
-           //  }
+            //} else {
+            //  $location.url(Routes.CONFIG);
+            //}
+
             $rootScope.$on('resume', function() {
+              if(configurationService.configurationComplete()) {
```
ericcf: This looks identical to line 89 above. What do you think about refactoring into a method?

src/js/Application.js
```diff
@@ -65,25 +65,59 @@
...
+              } else {
+                $location.url(Routes.HOME);
+              }
             } else {
               $location.url(Routes.SET_UP);
             }
-            // } else {
-            //   $location.url(Routes.CONFIG);
-           //  }
+            //} else {
```
ericcf: commented code

src/js/controllers/MorningEmaController.js
```diff
@@ -0,0 +1,70 @@
+(function() {
```
ericcf: This controller looks almost identical to the EveningEmaController. Is there a way to reuse the code?

src/js/controllers/SessionsController.js
```diff
@@ -61,6 +60,10 @@
     this.getCurrentSessionComplete = function() {
       return sessionsService.isComplete(sessionsService.currentSessionNumber());
     };
+
+    this.convertHtmlContent = function(rawContent) {
+      return $sce.trustAsHtml(rawContent);
```
ericcf: Why can't `rawContent` be sanitized instead?

src/js/controllers/MorningEmaController.js
```diff
@@ -0,0 +1,70 @@
...
+      var question;
+      if(sessionsService.cessationSession() === sessionsService.CESSATION_SESSION_PRE) {
+        question = preCessationQuestionState;
+      } else {
+        question = postCessationQuestionState;
+      }
+      return question;
+    };
+
+    this.appendToCurrentAnswer = function(valueToAppend) {
```
ericcf: Do you think this is an improvement or too terse?

```javascript
this.currentAnswer += (this.currentAnswer ? ', ' : '' + valueToAppend);
```

src/js/controllers/MorningEmaController.js
```diff
@@ -0,0 +1,70 @@
...
+        this.saveMorningEma();
+      }
+    };
+
+    this.saveAnswer = function(question, answer, nextQuestion) {
+      this.currentAnswer = answer;
+      this.saveQuestionAnswer(question, nextQuestion);
+    };
+
+    this.cessationSessionQuestionState = function(preCessationQuestionState,
```
ericcf: What about this?

```javascript
if(sessionsService.cessationSession() === sessionsService.CESSATION_SESSION_PRE) {
  return preCessationQuestionState;
}

return postCessationQuestionState;
```

src/js/services/Ema.js
```diff
@@ -0,0 +1,74 @@
...
+        }
+      }
+      return todaysMorningEma;
+    };
+
+    this.getTodaysEveningEma = function() {
+      var emaData = emaCache.fetchAllRaw();
+      var todaysEveningEma;
+      for (var i = 0; emaData && i < emaData.length; i++) {
+        if (moment(emaData[i].date) > moment().subtract(1, 'day')
```
ericcf: I'd suggest using the Moment Query methods instead: http://momentjs.com/docs/#/query/

Because you're comparing objects, `<`, `>`, etc. may not be well defined.

src/js/controllers/SessionsController.js
eschlange: I'm not sure what you mean by this.

src/js/controllers/MorningEmaController.js
eschlange: Yes, at the very least I'll try to abstract out the common components.

src/js/controllers/SessionsController.js
ericcf: My understanding is that `trustAsHtml` blindly marks the entire string as safe, whereas `sanitize` filters out anything not in a whitelist.

src/js/controllers/MorningEmaController.js
eschlange: Personally, I don't like multiple return statements in a method, but this is viable code. Would you like me to change it?

src/js/services/Ema.js
eschlange: Cool! I'll check these out.

src/js/controllers/MorningEmaController.js
ericcf: That's fair. It's fine by me if you want to keep as is.

test/unit/controllers/RiskyTimesControllerSpec.js
```diff
@@ -2,74 +2,92 @@
...
+    $provide.service('configurationService', configurationServiceMock);
+  }));
+
+  beforeEach(module('sis.controllers'));
+
+  beforeEach(inject(function ($controller) {
+    controller = $controller('RiskyTimesController');
+  }));
+
+  describe('#translateDay', function () {
```
ericcf: I don't understand these specs. It looks like `translateDay` doesn't take any arguments.

test/unit/controllers/RiskyTimesControllerSpec.js
```diff
@@ -2,74 +2,92 @@
...
-            expect(controller.riskyTimeTime).to.not.equal(undefined);
-        });
-        it('populates errors when a description is not set', function() {
-            controller.riskyTimeDay = 'Monday';
-            controller.riskyTimeTime = new Date();
-            controller.saveRiskyTime();
-            expect(controller.riskyTimeDay).to.not.equal(undefined);
-        });
+  describe('#saveRiskyTime', function () {
+    it('populates errors when a time is not set', function () {
```
ericcf: The only place I can find where `riskyTimeDay` gets set to something other than `''` is in the `selectWeekday` method. What does `controller.riskyTimeDay !== undefined` tell us? (Maybe I misunderstand what's going on.)

test/unit/controllers/RiskyTimesControllerSpec.js
```diff
@@ -2,74 +2,92 @@
...
 
-        it('should save a new, valid risky time', function() {
-            controller.riskyTimeDay = 'Monday';
-            controller.riskyTimeTime = new Date();
-            controller.riskyTimeDescription = 'Some description.';
-            controller.saveRiskyTime();
-            expect(controller.riskyTime.day).to.equal(undefined);
-            expect(controller.riskyTime.time).to.equal(undefined);
-        });
+    it('should save a new, valid risky time', function () {
```
ericcf: How do these expectations show that a new valid risky time is being saved?

src/partials/common_ema_questions.html
```diff
@@ -0,0 +1,435 @@
+
+<!-- START STATE_CONFIDENT_WEEKS_PRE (PRE-QUIT) -->
+<div ng-show="ema.currentQuestion == ema.questions.STATE_CONFIDENT_WEEKS_PRE">
+    <h4>How CONFIDENT are you that you will be able to quit smoking for at least 2 WEEKS?</h4>
+    <select class="form-control"
```
ericcf: The next time you need to add/change something on this form, I'd suggest refactoring into directives to make it easier to update. It looks like there are a lot of similar form fields.

src/js/Application.js
```diff
@@ -65,35 +64,52 @@
           templateUrl: 'partials/session.html',
           controller: 'SessionsController',
           controllerAs: 'session'
+        })
+        .when(Routes.MORNING_EMA, {
+          templateUrl: 'partials/morning.html',
+          controller: '' +
+          'EmaController',
```
michaelwehrley: just to be obnoxiously curious - how come this spans 2 lines?

eschlange: Not sure, but it's probably either a mistaken carriage return or IDE auto-formatting.

test/unit/services/EmaSpec.js
```diff
@@ -0,0 +1,91 @@
...
+      expect(emaService.isMorningPeriod(moment().hours(20))).to.equal(false);
+    });
+  });
+
+  describe('#isEveningPeriod', function () {
+    it('should return true if time passed in is within params of evening time.', function () {
+      expect(emaService.isEveningPeriod(moment().hours(20))).to.equal(true);
+    });
+    it('should return true if time passed in is within params of evening time.', function () {
+      expect(emaService.isEveningPeriod(moment().hours(10))).to.equal(false);
```
michaelwehrley: I think you mean "should return false".

test/unit/controllers/RiskyTimesControllerSpec.js
```diff
@@ -2,74 +2,92 @@
...
-        });
+  describe('#saveRiskyTime', function () {
+    it('populates errors when a time is not set', function () {
+      controller.riskyTimeDay = 'Monday';
+      controller.riskyTimeDescription = 'Some description.';
+      controller.saveRiskyTime();
+      expect(controller.riskyTimeDay).to.not.equal(undefined);
+    });
+    it('populates errors when a day is not set', function () {
+      controller.riskyTimeTime = new Date();
```
michaelwehrley: For the future, it might be easier to understand if you rename it to riskyTimeDateTime.

src/partials/morning.html
```diff
@@ -0,0 +1,67 @@
...
+        <!-- END STATE_TIME_AWAKE -->
+
+        <!-- START STATE_SLEEP_QUALITY -->
+        <div ng-show="ema.currentQuestion == ema.questions.STATE_SLEEP_QUALITY">
+            <h4>How well did you sleep last night?</h4>
+            <select class="form-control"
+                    ng-model="ema.currentAnswer"
+                    ng-change="ema.saveQuestionAnswer('How well did you sleep last night?',
+                                                      ema.questions.STATE_RESTED)">
+                <option ng-repeat="(index, value) in ['0 - not well at all', '1','2','3','4','5 - somewhat well','6','7','8','9','10 - extremely well']"
```
michaelwehrley: It is kind of funny to see all this in the view.  It totally works, but I think of this usually in a helper in rails - not sure where it would go in angular.  A service maybe?

src/partials/common_ema_questions.html
```diff
@@ -0,0 +1,435 @@
...
+                                        'stranger']">
+        <input type="checkbox" ng-model="index" ng-change="ema.appendToCurrentAnswer(value)" value="{{value}}" name="company"> {{ value }}<br>
+    </div>
+    <button ng-show="ema.currentAnswer" class="btn btn-primary" ng-click="ema.saveAnswer('In the last 15 minutes, have you seen any of these people smoke? (check all that apply)', ema.currentAnswer, ema.questions.STATE_LOCATION)">Continue</button>
+</div>
+<!-- END STATE_CONTEXT_COMPANY_SMOKE -->
+
+<!-- START STATE_LOCATION -->
+<div ng-show="ema.currentQuestion == ema.questions.STATE_LOCATION">
+    <h4>Where are you at the moment?</h4>
```
michaelwehrley: Reasonable to use `ng-repeat` here?

src/partials/common_ema_questions.html
```diff
@@ -0,0 +1,435 @@
...
+                                        'other person you know',
+                                        'stranger']">
+        <input type="checkbox" ng-model="index" ng-change="ema.appendToCurrentAnswer(value)" value="{{value}}" name="company"> {{ value }}<br>
+    </div>
+    <button ng-show="ema.currentAnswer" class="btn btn-primary" ng-click="ema.saveAnswer('In the last 15 minutes, have you seen any of these people smoke? (check all that apply)', ema.currentAnswer, ema.questions.STATE_LOCATION)">Continue</button>
+</div>
+<!-- END STATE_CONTEXT_COMPANY_SMOKE -->
+
+<!-- START STATE_LOCATION -->
+<div ng-show="ema.currentQuestion == ema.questions.STATE_LOCATION">
```
michaelwehrley: It looks like you are comparing two strings here.  `===` instead?

src/js/services/LocalNotifications.js
```diff
@@ -0,0 +1,25 @@
...
+    this.schedule = function(notification) {
+      $window.plugin.notification.local.schedule(notification);
+    };
+
+    this.clearAllTriggered = function() {
+      $window.plugin.notification.local.clearAll();
+    };
+
+    this.isPresent = function(notificationId, callback) {
+      $window.plugin.notification.local.isPresent(notificationId, callback);
```
michaelwehrley: Did you get this [`notification.local.isPresent`] to work?!

src/js/services/Ema.js
```diff
@@ -0,0 +1,81 @@
...
+          todaysMorningEma = emaData[i];
+          break;
+        }
+      }
+      return todaysMorningEma;
+    };
+
+    this.getTodaysEveningEma = function() {
+      /*eslint-disable */
+      debugger;
```
michaelwehrley: :smile_cat:

src/js/services/Ema.js
```diff
@@ -0,0 +1,81 @@
...
+      /*eslint-enable */
+      var emaData = emaCache.fetchAllRaw();
+      var todaysEveningEma;
+      for (var i = 0; emaData && i < emaData.length; i++) {
+        if (moment(emaData[i].date).local() > moment().subtract(1, 'day')
+            .hours(this.EMA_EVENING_START_HOUR).minutes(0).seconds(0) &&
```
michaelwehrley: You might be able to use `moment().startOf('hour')` to round down to the nearest hours.

src/js/services/Ema.js
```diff
@@ -0,0 +1,81 @@
...
+      return todaysMorningEma;
+    };
+
+    this.getTodaysEveningEma = function() {
+      /*eslint-disable */
+      debugger;
+      /*eslint-enable */
+      var emaData = emaCache.fetchAllRaw();
+      var todaysEveningEma;
+      for (var i = 0; emaData && i < emaData.length; i++) {
```
michaelwehrley: This is probably me, but I am confused here.  if `emaData` is [], would it fail?  Do you instead mean `(var i = 0; emaData.length > 0 && i < emaData.length; i++)`

src/js/services/Ema.js
```diff
@@ -0,0 +1,81 @@
...
+    this.EMA_MORNING_START_HOUR = 9;
+    this.EMA_EVENING_START_HOUR = 19;
+
+    this.saveEma = function (emaQuestions, emaPeriod) {
+      var ema = {};
+      ema.questions = emaQuestions;
+      ema.type = emaPeriod;
+      ema.date = moment().local().toJSON();
+      /*eslint-disable */
+      debugger;
```
michaelwehrley: note.

src/js/controllers/SessionsController.js
eschlange: Eric, are you referring to the explanation found under the header "This feels like too much overhead" at https://docs.angularjs.org/api/ng/service/$sce. In other words, should I include the ngSanitize module along with my current solution?

test/unit/controllers/RiskyTimesControllerSpec.js
eschlange: That is correct, `translateDay` pulls the value from the parent function, set in real time through the view.

src/js/controllers/SessionsController.js
ericcf: I'm just asking the question: if you can sanitize rather than trust and it still works, wouldn't that be a safer approach? And if you can't, why not?

test/unit/controllers/RiskyTimesControllerSpec.js
ericcf: So each spec below collapses to:

```javascript
expect(controller.translateDay())
  .to.equal(controller.riskyTime.day);
```

?

test/unit/controllers/RiskyTimesControllerSpec.js
eschlange: Good catch! I'll update to check for an empty string rather than undefined.

test/unit/controllers/RiskyTimesControllerSpec.js
eschlange: I'll update the description to better describe this test.

src/partials/common_ema_questions.html
eschlange: I may come to you in the future in order to get some advice. This sounds like an ideal solution for this scenario, but I'm not sure I can fit this into a refactor right now. Especially since the questions may potentially change and some may need to be customized.

test/unit/services/EmaSpec.js
eschlange: good catch!

test/unit/controllers/RiskyTimesControllerSpec.js
eschlange: Noted, I'll update this when I make further revisions to this file in upcoming stories regarding the merging of risky times and strategies.

src/partials/morning.html
eschlange: I'll likely try to create some directives for these and move the data over to the controller or service layer. As I commented with regards to Eric's mention of this, I'm not sure I have time for a refactor at this point, especially if these questions and their flow may change before initial release.

src/partials/common_ema_questions.html
eschlange: No, not in my opinion. I'd be abstracting data into some sort of two dimensional array. Perhaps if I were to refactor this into a directive, but this is the only case of this sort of question.

src/partials/common_ema_questions.html
eschlange: I've had issues in the past using the triple equals sign within the context of an angular directive parameter. Is what you mention even possible?

src/js/services/LocalNotifications.js
eschlange: (Yes I did get it working.) In this case though, this method is left as a placeholder for the upcoming notification logic.  I'm not sure why isPresent was not working when we work working together on debugging this. It turns out that isPresent was working, but not in the scope of the debugger's context. If I repeatedly ran the method in the console, the callback eventually runs.

src/js/services/Ema.js
eschlange: Removed, was in the middle of debugging a newly introduced error last night.

src/js/services/Ema.js
eschlange: De-debuggered.

src/js/controllers/SessionsController.js
eschlange: I think that in order to answer your question, I need a better understanding of what the standard way to sanitize data in Angular (I'm not sure what this is.) If I knew what you meant by sanitize, I could give you a more accurate answer on why I'd use one over the other. 

I used this particular approach because it seemed to be a common approach on StackOverflow.

ericcf: The only way I'm aware of to sanitize content is with `$sanitize`: https://docs.angularjs.org/api/ngSanitize/service/$sanitize, so I guess it's the standard?

test/unit/controllers/RiskyTimesControllerSpec.js
eschlange: Ah I see, fixing the specs now.

src/js/services/Ema.js
eschlange: so change this: ```moment(emaData[i].date).local() > moment().subtract(1, 'day').hours(this.EMA_EVENING_START_HOUR).minutes(0).seconds(0)``` to this: ```moment(emaData[i].date).local() > moment().subtract(1, 'day')
+            .hours(this.EMA_EVENING_START_HOUR).startOf('hour')```?

michaelwehrley: That would be right.  Not sure that helps too much though.

src/partials/common_ema_questions.html
michaelwehrley: I did it here: https://github.com/cbitstech/ed_navigator_app/blob/master/src/partials/directives/categoryThumbnail.html#L1.

src/js/services/Ema.js
eschlange: I like it: less verbose, more accurate.

src/js/controllers/SessionsController.js
eschlange: In this case, all of the labels being trusted as HTML content is coming from a static file/JSON object within the application. Would there ever be a case in this scenario where trusting this content wouldn't be safe?

ericcf: Maybe - how is the JSON being generated?

eschlange: The JSON is generated from a static file that defines it: SessionContent.js. (https://github.com/cbitstech/sis_app/blob/ems-95522128-morning-evening-notifications/src/js/constants/SessionContent.js)

eschlange:
@ericcf @michaelwehrley 
I think I covered all of the refactoring based on your comments. If one of you has a moment, would you please merge this in? I'd like to get this to Chris for testing sometime soon.

src/js/Application.js
ericcf: Looks like you should be able to just replace it with `'EmaController'`?

eschlange: Ah yes, done.

test/unit/controllers/RiskyTimesControllerSpec.js
ericcf: Is this example description correct? I don't see anything in the test body related to errors.

eschlange: Updated the test descriptions and the test contents to accurately describe what is being tested.

ericcf:
Thanks, I think our comments have all been addressed!

PR #24, opened 2015-08-10
Add completion notice on home page for complete sessions

eschlange:
- Updated view to include notification of completed sessions on home page.
- Used existing session service methods to accomplish this.
- Added session service method mock to home controller test.

[#95522334]


src/js/controllers/HomeController.js
```diff
@@ -21,6 +21,10 @@
     this.cessationDayOffsetValue = undefined;
     this.cessationDayOffsetAbsValue = undefined;
 
+    this.sessionOneComplete = sessionsService.isComplete(1);
```
ericcf: What do you think about delegating instead here, e.g.

```javascript
this.isSessionComplete = function(session) {
  return sessionsService.isComplete(session);
}
```

My thought is that it reduces the controller API and makes it more flexible (in the event that more sessions are added).

ericcf:
Maybe these two commits could be combined? It would probably make it easier to read the commit history in the future.

src/js/controllers/HomeController.js
eschlange: I like your idea. The only reason I didn't do that originally was because I've had issues in the past with setting Angular conditionals to equal service functions. Is there a reason why Angular views sometimes break when conditional view logic is set to a service function call?

I made the change you suggested and it is working correctly.

eschlange:
I'll try a git squash next time I run into this scenario.

eschlange:
@ericcf I updated the controller with your suggestion. Its much cleaner now!

ericcf:
:ocean:

PR #25, opened 2015-08-10
Increase daily exercise textarea height.

eschlange:
- Also removed unnecessary code (for another view) from the exercise view.

[#98735292]


ericcf:
Deleting code :metal:

PR #26, opened 2015-08-10
Remove risky time strategies.

eschlange:
- Removed view.
- Removed configuration logic to ensure risky time strategies exist.
- Removed from routes.

[#99282286]


ericcf:
:eyes:

PR #29, opened 2015-08-13
Disable next button for sessions when unanswered.

eschlange:
[#101203814]


src/js/controllers/SessionsController.js
```diff
@@ -20,6 +20,7 @@
         question: this.currentQuestion,
         answer: this.currentAnswer
       });
+      this.currentAnswer = undefined;
```
ericcf: Why did you decide to initialize to `undefined` instead of `null`?

michaelwehrley: How come you need to specify that it is `undefined`?

test/unit/controllers/SessionsControllerSpec.js
```diff
@@ -114,4 +114,22 @@ describe('SessionsController', function () {
       expect(controller.getSessionContent()).to.equal('some_content');
     });
   });
+
+  describe('#disableNext', function() {
```
ericcf: This spec looks good, but there are no specs for `#saveSessionAnswer` or `#next` on this controller.

eschlange:
@ericcf @michaelwehrley Updated the controller to check for nulls, didn't know that it was bad practice to use undefined until now. 

Also added specs for some of the older methods.

ericcf:
One argument against initializing a property to `undefined` is that it becomes ambiguous as to whether the property is uninitialized or initialized to `undefined`:

```javascript
a = {};
a.foo = undefined;
a.foo === undefined; // true
b = {};
b.foo === undefined; // true
a.foo === b.foo; // true
```

Another thing to note:

```javascript
a = {};
a.foo === undefined; // true
a.foo == null; // true
a.foo === null; // false !
```

michaelwehrley:
I was wondering if that was part of the argument.  Thanks for the example and clarification.

PR #30, opened 2015-08-17
Add continue button for EMA questions, override back button.

eschlange:
* Ema questions now require the press of a 'continue' button.
* EMA and session pages no longer allow the 'back button' navigation.
* Added specs for new methods and updated existing test cases.

[#101067586]


test/unit/controllers/EmaControllerSpec.js
```diff
@@ -88,13 +94,24 @@ describe('EmaController', function() {
...
     });
 
     it('should append the given answer with formatting to the controllers current answer.', function() {
-      controller.appendToCurrentAnswer('an answer');
-      controller.appendToCurrentAnswer('a second answer');
+      var questionText = 'How well did you sleep last night?';
+      var firstAnswer = 'an answer';
+      var secondAnswer = 'a second answer';
+
+      controller.appendToCurrentAnswer(questionText, firstAnswer, controller.questions.STATE_CAFFEINE);
```
ericcf: Why do you pass in values for the first and third arguments?

src/js/controllers/EmaController.js
```diff
@@ -54,13 +61,21 @@
       return question;
     };
 
-    this.appendToCurrentAnswer = function(valueToAppend) {
+    this.appendToCurrentAnswer = function(question, valueToAppend, nextQuestion) {
+      this.currentQuestionText = question;
```
ericcf: Does it make sense to replace these two lines with `this.saveQuestionAnswer(question, nextQuestion);`?

src/js/controllers/EmaController.js
```diff
@@ -54,13 +61,21 @@
...
+      this.nextQuestion = nextQuestion;
+
       if(this.currentAnswer) {
         this.currentAnswer += ', ' + valueToAppend;
       } else {
         this.currentAnswer = valueToAppend;
       }
     };
+
+    this.singleClickSaveAnswer = function(question, answer, nextQuestion) {
```
ericcf: Are you still using the `saveAnswer` method?

test/unit/controllers/EmaControllerSpec.js
eschlange: I use this method signature in order to ensure that the question text and next question fields are set in all cases.

src/js/controllers/EmaController.js
eschlange: this.saveQuestionAnswer would override answers in the case of checkboxes.

src/index.html
```diff
@@ -90,6 +91,19 @@
                 $rootScope.$broadcast('resume');
             }
         }
+
+        function handleBackButton() {
```
ericcf: I think this logic will be brittle because it doesn't have tests and references strings rather than named variables.

src/js/controllers/EmaController.js
eschlange: This is used for cases where the answer is given through a button press.

src/js/controllers/EmaController.js
ericcf: I'm not sure I understand. The body of `saveQuestionAnswer` is identical to these two lines:
```javascript
this.currentQuestionText = question;
this.nextQuestion = nextQuestion;
```

test/unit/controllers/EmaControllerSpec.js
ericcf: I meant in the test specifically. If you called `controller.appendToCurrentAnswer(null, firstAnswer, null);` the test would still pass right?

src/index.html
eschlange: I agree with you on this one. I'll abstract out the back button handler and write a spec for the method.

src/js/controllers/EmaController.js
```diff
@@ -54,13 +61,21 @@
       return question;
     };
 
-    this.appendToCurrentAnswer = function(valueToAppend) {
+    this.appendToCurrentAnswer = function(question, valueToAppend, nextQuestion) {
```
michaelwehrley: It would be helpful to change this parameter from `question` to `questionText` since that is what it is and not the question object.

src/partials/common_ema_questions.html
```diff
@@ -327,22 +326,21 @@
...
-    <button class="ema-button btn btn-primary" ng-click="ema.saveAnswer('Where are you at the moment?', 'school', ema.questions.STATE_LOCATION_SCHOOL)">school</button><br>
-    <button class="ema-button btn btn-primary" ng-click="ema.saveAnswer('Where are you at the moment?', 'work', ema.questions.STATE_LOCATION_WORK)">work</button><br>
-    <button class="ema-button btn btn-primary" ng-click="ema.saveAnswer('Where are you at the moment?', 'bar / restaurant / party', ema.questions.STATE_SUBSTANCES)">bar / restaurant / party</button><br>
-    <button class="ema-button btn btn-primary" ng-click="ema.saveAnswer('Where are you at the moment?', 'in transition / traveling', ema.questions.STATE_LOCATION_TRAVEL)">in transition / traveling</button><br>
-    <button class="ema-button btn btn-primary" ng-click="ema.saveAnswer('Where are you at the moment?', 'other', ema.questions.STATE_SUBSTANCES)">other</button>
+    <button class="ema-button btn btn-primary" ng-click="ema.singleClickSaveAnswer('Where are you at the moment?', 'home', ema.questions.STATE_SUBSTANCES)">home</button><br>
```
michaelwehrley: Would it make sense to iterate through an array?
```
<div ng-repeat="prompt in ema.questionPrompts()">
  <button class="ema-button btn btn-primary"
          ng-click="ema.singleClickSaveAnswer(
                     'Where are you at the moment?'',
                     prompt.location,
                     prompt.questionState)">
  prompt.text
</button><br>
</div>
````

File: test/unit/controllers/EmaControllerSpec.js [hunk 7]
```diff
@@ -88,13 +94,24 @@ describe('EmaController', function() {
 
   describe('#appendToCurrentAnswer', function() {
     it('should append the given answer to the controllers current answer.', function() {
-      controller.appendToCurrentAnswer('an answer');
-      expect(controller.currentAnswer).to.equal('an answer');
+      var questionText = 'How well did you sleep last night?';
+      var answer = 'an answer'
+
```
michaelwehrley: would it be helpful to assert before hand?
```
expect(controller.currentAnswer).to.not.equal(answer);
expect(controller.nextQuestion).to.not.equal(controller.questions.STATE_CAFFEINE);
expect(controller.currentQuestionText).to.not.equal(questionText);
```

src/partials/common_ema_questions.html
eschlange: It would need to be a two dimensional array due to the following questions being different in several cases. This current solution seems less complex to me.

src/js/controllers/EmaController.js
eschlange: Ah, I misunderstood. Yes, this is redundant. I'll do as you suggest.

test/unit/controllers/EmaControllerSpec.js
eschlange: That's correct. I'll remove those lines since they aren't pertinent to this test.

test/unit/controllers/EmaControllerSpec.js
eschlange: I don't know. Doing that would imply that there may be a chance for the initialized state of the controller may contain the same values as the test input. It seems gratuitous in this case.

src/index.html
eschlange: I abstracted this method out into a separate function and utilized constants that represent the urls, but writing tests for this is not a simple matter. I just sunk about an hour into trying to use our test framework to do this, and found that Sinon cannot directly overwrite 'window.location' methods or variables (http://stackoverflow.com/a/7050884/512463). The solutions that I found to get around this involve writing wrappers for these window calls, then mocking those. 

This is a very small piece of code, I don't think I should waste any more time trying to write a test for it.

ericcf: It's pretty straightforward to mock out the `$window` service in Angular. See for example https://github.com/cbitstech/focus_app/blob/ecf_101161546_extra_prompt_response/test/unit/controllers/SlideControllerSpec.js, https://github.com/cbitstech/medlink2_app/blob/8ad04e26b4e5ea3a715913a38737d5db6a318c6a/test/unit/services/ExceptionReportSpec.js, and https://github.com/cbitstech/medlink2_app/blob/ba435fa89561599f4d8e013a53e65bbf6dc0d822/test/unit/resources/SettingsSpec.js

Angular's dependency injection and the fact that it wraps any browser global object means that you can mock out just about anything.

eschlange: In your case, you are adding the mock through a controller. In my case, I'm outside of the context of a controller or service. Do you have any recommendations on how I can wrap this into an Angular context if I'm adding the listener within initialization of the Angular application?

src/js/controllers/EmaController.js
eschlange: I agree, making the change now.

src/index.html
ericcf: I linked to one controller spec and two service specs.

eschlange:
@michaelwehrley @ericcf Unless one of you has further comments, I believe addressed all of the mentioned issues.

src/js/controllers/EmaController.js
michaelwehrley: I think what @ericcf means, it it looks like `saveAnswer` was changed to `singleClickSaveAnswer`.  Thus are you still using `saveAnswer` or should it be `singleClickSaveAnswer`?

michaelwehrley:
It looks like you addressed all comments except for maybe https://github.com/cbitstech/sis_app/pull/30#discussion-diff-37206323?  Otherwise, it looks good! Great work.

eschlange:
I see it now. I must have been in the middle of merging those two methods on a previous merge. I merged the two methods and updated my unit tests for the controller.

michaelwehrley:
:shipit:

PR #34, opened 2015-08-20
Add strategy selection to risky time edits.

eschlange:
* Update styles for risky time and edit pages.
* Created view for editing risky time strategies.
* Added bootstrap ui angular library.
* Added a spec file and tests for the new editing controller.
* Updated configuration files with new files.

[#99282266]


src/js/Application.js
```diff
@@ -113,12 +113,12 @@
...
       ['ngResource', 'sis.services', 'sis.constants']);
   angular.module('sis.controllers',
-      ['sis.constants', 'sis.resources', 'sis.services']);
+      ['sis.constants', 'sis.resources', 'sis.services', 'ui.bootstrap']);
   angular.module('sis.services',
       ['sis.resources', 'sis.constants']);
   angular.module('sis',
       ['ngRoute', 'mobiscroll-datetime', 'sis.controllers',
-        'sis.resources', 'sis.services', 'sis.constants'])
+        'sis.resources', 'sis.services', 'sis.constants', 'ui.bootstrap'])
```
michaelwehrley: Does this mean you are accessing this module in other places then the controllers?

test/unit/controllers/RiskyTimesEditControllerSpec.js
```diff
@@ -0,0 +1,64 @@
...
+  beforeEach(module('sis.controllers'));
+
+  beforeEach(inject(function ($controller) {
+    controller = $controller('RiskyTimesEditController');
+  }));
+
+  describe('#updateRiskyTimeStrategy', function() {
+    it('should use the configuration service to save the updated risky time.', function () {
+      controller.updateRiskyTimeStrategy();
+      expect(configurationServiceSaveRiskyTimesSpy.calledOnce).to.equal(true);
```
ericcf: What do you think about adding an expectation for the method argument as well?

src/js/controllers/RiskyTimesController.js
```diff
@@ -1,7 +1,7 @@
 (function () {
   'use strict';
 
-  function RiskyTimesController(configurationService) {
+  function RiskyTimesController(configurationService, sessionsService, $modal) {
```
ericcf: It looks like you forgot to update the specs for this controller.

src/js/Application.js
eschlange: Good catch, that is unnecessary. Remnants of an older implementation. Removing it now.

test/unit/controllers/RiskyTimesEditControllerSpec.js
eschlange: I think that is a good idea, especially since it is looking for a very structured argument. I'll add it to the spec now.

src/js/controllers/RiskyTimesController.js
eschlange: Good catch! I'm updating method signature and adding messing tests for new method.

src/js/controllers/RiskyTimesController.js
```diff
@@ -1,7 +1,7 @@
...
     this.RISKY_TIMES_KEY = 'risky_times_key';
 
     this.SUNDAY_KEY = 'su';
```
michaelwehrley: I know this is a bit late and theoretical, but it might make sense for these constants to be placed in an `abbreviatedDay` service.  Then when you reference them, you could call   `abbreviatedDay.sunday()` etc.  Just a thought.

src/js/controllers/RiskyTimesEditController.js
```diff
@@ -0,0 +1,36 @@
...
+      this.riskyTimes.splice(this.riskyTimeIndex, 0, this.riskyTime);
+      configurationService.saveRiskyTimes(this.riskyTimes);
+      $modalInstance.close();
+    };
+
+    this.cancel = function () {
+      $modalInstance.close();
+    };
+
+
```
michaelwehrley: Does jslint look for these extra lines?

ericcf: @michaelwehrley FYI we're using eslint, not jslint. Here are the configurable rules: http://eslint.org/docs/rules/

src/partials/risky_times.html
```diff
@@ -49,3 +55,4 @@
 <script type="text/javascript">
   $("#risky_time_time").mobiscroll().time();
```
michaelwehrley: I can't seem to find this on the page?  Did you mean `$("[name=risky_time_time]")` or `$("[input=risky_time_time]")`

src/js/controllers/RiskyTimesEditController.js
michaelwehrley: Nice. Food for thought: http://eslint.org/docs/rules/no-multiple-empty-lines.

ericcf: @michaelwehrley here's the rule that would handle that: http://eslint.org/docs/rules/no-multiple-empty-lines

I think we should start building up a standard `.eslintrc` file to use across projects for consistency. Maybe we create a cbitstech `dotfiles` repo for it and other config files? @eschlange

src/js/controllers/RiskyTimesController.js
eschlange: If the date manipulation logic bleeds out of this controller, I'll abstract it into its own service. Currently, its only being used in this one location and I don't see it being used elsewhere.

src/js/controllers/RiskyTimesEditController.js
eschlange: I added the extraneous blank line rule to my eslint config. I also created a Confluence page that lists references to our existing eslint project configs (https://confluence.nubic.northwestern.edu/display/~ems408/eslint+configurations.) I think if we plan to do several more projects that require it, there are benefits to creating a config repository. We'll probably want to have a short discussion about the various configurations we'd like to add to the repos. Otherwise we can covert the eslint Confluence page to apply to cover additional configuration files from Github.

michaelwehrley: :+1: thanks!

src/partials/risky_times.html
eschlange: That is bizarre, considering that the mobilescroll functionality is still being triggered properly. I'll add the missing ID to the applicable tag.

eschlange:
@ericcf @michaelwehrley Fixes are in place where applicable, based on your comments. Thanks guys!

ericcf:
Looks pretty good to me.

michaelwehrley:
:+1: :rocket:

PR #36, opened 2015-08-25
Add risky time notifications and motivational view.

eschlange:
- Added risky time strategy selection and view.
- Added new risky time notifications.
- Updated configuration files to include new controllers.
- Added spec coverage for new controllers.
- Added spec coverage to existing controllers.

[#99282204]


src/js/controllers/RiskyTimesEditController.js
```diff
@@ -11,6 +11,26 @@
...
+    this.riskyTimeNotificationId = riskyTimeToEdit.notificationId;
+
+    this.updateNotification = function(riskyTime) {
+      var NOTIFICATION_TEXT_MAX_SIZE = 100;
+      var riskyDateTime = new Date(riskyTime.time);
+      return {
+        id: riskyTime.notificationId,
+        title: 'Your strategy',
+        text: riskyTime.strategy.substr(0, NOTIFICATION_TEXT_MAX_SIZE),
+        firstAt: $window
```
ericcf: Does this method chain (`...toDate()`) return a `Date` object with a different value than `riskyDateTime`?

src/js/controllers/MotivationController.js
```diff
@@ -0,0 +1,34 @@
+(function () {
+  'use strict';
+
+  function MotivationController(configurationService, $routeParams) {
```
ericcf: Looks like this needs unit tests.

src/js/controllers/RiskyTimesEditController.js
eschlange: Yes, and that is preferred. If I didn't do that, the date would be in the past, as opposed to the current or next day that a notification was to trigger. If it was in the past, the notification would trigger immediately.

src/js/controllers/MotivationController.js
eschlange: On it~

test/unit/controllers/RiskyTimesControllerSpec.js
```diff
@@ -103,7 +153,51 @@ describe('RiskyTimesController', function () {
...
+
+    it('gets the risky times from the controller if it is set.', function () {
+      controller.riskyTimes = RISKY_TIMES;
+      controller.getRiskyTimes();
+      expect(controller.riskyTimes).to.equal(RISKY_TIMES);
+      expect(configurationServiceGetRiskyTimesSpy.calledOnce).to.equal(false);
+    });
+  });
+
+  describe('#randomNotification', function () {
```
ericcf: It might be a good idea to cover each notification type scenario. Looks like this only covers one.

eschlange: I agree, I'll mock out the random methods and write test cases for the other two options.

src/js/controllers/RiskyTimesController.js
```diff
@@ -64,14 +69,93 @@
...
+      var NOTIFICATION_TEXT_MAX_SIZE = 100;
+      var MAX_RANDOM_TYPE_PRE_SESSION_2 = 2;
+
+      var notificationType;
+
+      if(riskyTime.strategy) {
+        notificationType = NOTIFICATION_TYPE_3;
+      } else {
+        if(this.cessationReasons) {
+          notificationType = Math.floor(Math.random() * MAX_RANDOM_TYPE_PRE_SESSION_2);
```
ericcf: Wrapping `Math.random()` with a service might be worth it so you could make your tests deterministic.

test/unit/controllers/RiskyTimesControllerSpec.js
```diff
@@ -16,17 +25,58 @@ describe('RiskyTimesController', function () {
...
     };
 
     $modalMock = function () {
-      this.open
+      this.open = function() {
+      };
+      $modalOpenSpy = sinon.spy(this, 'open');
+    };
+
+    $windowMock = function () {
```
ericcf: Remember, you can generally provide mock services as constants instead, and just pass in an object instead of a generator function.

ericcf: Remember, you can generally provide mock services as constants instead, and just pass in an object instead of a generator function.

test/unit/controllers/RiskyTimesControllerSpec.js
```diff
@@ -16,17 +25,58 @@ describe('RiskyTimesController', function () {
...
+          day: empty,
+          hours: empty,
+          minutes: empty,
+          toDate: empty
+      };
+      this.moment = function () {
+        return chainedMomentMethods;
+      };
+      $windowMomentSpy = sinon.spy(this, 'moment');
+      sinon.stub(chainedMomentMethods, 'local').returns(chainedMomentMethods);
```
ericcf: :eyes:

test/unit/controllers/RiskyTimesControllerSpec.js
```diff
@@ -103,7 +153,51 @@ describe('RiskyTimesController', function () {
...
+  });
+
+  describe('#getRiskyTimes', function () {
+    it('gets the risky times using the configuration service if it isn\'t already set.', function () {
+      controller.getRiskyTimes();
+      expect(controller.riskyTimes).to.equal(RISKY_TIMES);
+      expect(configurationServiceGetRiskyTimesSpy.calledOnce).to.equal(true);
+    });
+
+    it('gets the risky times from the controller if it is set.', function () {
```
ericcf: This test ends up being identical to the one above it. Would it be worth changing the first line to something like this?

```javascript
var alternativeRiskyTimes = ['foo', 'bar'];
controller.riskyTimes = alternativeRiskyTimes;
```

src/js/controllers/RiskyTimesController.js
eschlange: I abstracted the notification building out to a separate function in order to test the remaining notification. It seems overkill to generate that much code just to test a single Javascript statement that is mostly native.

test/unit/controllers/RiskyTimesControllerSpec.js
eschlange: Sinon doesn't offer anything that abstracts out chainedMethods for testing purposes. Unfortunately, this is one of the cleanest options I found.

test/unit/controllers/RiskyTimesControllerSpec.js
eschlange: Very good point. I'll add in the variety for the risky times.

test/unit/controllers/RiskyTimesControllerSpec.js
eschlange: Yes, that was extremely helpful for a few of my previous tests. I'll definitely be using it in the future when the need arises.

eschlange:
@ericcf Thanks for the comments! My latest changes include updates based on some of your comments.

ericcf:
:mailbox: :email:

PR #41, opened 2015-08-27
Prepare application for Ruby-Appium testing.

eschlange:
* Added gemfile for necessary Ruby gems.
* Add configuration to start and run appium tests.
* Add same tests and appium config.

[#101508662]


test/features/android.rb
```diff
@@ -0,0 +1,51 @@
+require "rubygems"
+require "appium_lib"
+require "rspec"
+
+caps_file = File.join(File.dirname(__FILE__), "appium.txt")
+caps = Appium.load_appium_txt(file: caps_file)
+
+RSpec.describe "configuration content" do
+  driver = nil
```
ericcf: I'd refactor line 9-17 and move it outside the `describe` block:

```ruby
RSpec.configure do |config|
  config.before(:example) do
    driver = Appium::Driver.new(caps).start_driver
    Appium.promote_appium_methods RSpec::Core::ExampleGroup
  end

  config.after(:example) do
    driver_quit
  end
end
```

See https://relishapp.com/rspec/rspec-core/v/3-3/docs/hooks/before-and-after-hooks#define-`before`-and-`after`-blocks-in-configuration

You may want to try before and after `context` or `suite` for efficiency. It's going to get really slow to start and stop with each example probably (I could be wrong).

eschlange: Thanks, Eric. I was looking for something to speed that up. I'll rework this now.

test/features/appium.txt
```diff
@@ -0,0 +1,9 @@
+[caps]
+platformName = "Android"
+platformVersion = "23.0.0"
```
ericcf: This line may not matter. The example I tried worked with `platformVersion = ""`. YMMV

eschlange:
@ericcf I abstracted out the before and after blocks by using :context. Not sure if it runs faster, but it definitely cleans up the code.

PR #42, opened 2015-09-01
Remove redundant dependency, added missing.

eschlange:
[#98948048]


eschlange:
@ericcf @cbrenner04  I removed the redundant dependency and added a missing local notifications plug-in. I created a clean slate and successfully ran an Android build. Would you please try building again to see if you receive any errors?

package.json
```diff
@@ -14,7 +14,8 @@
     "angular-resource": "^1.4.0",
     "angular-route": "^1.4.0",
     "cordova": "^5.0.0",
-    "moment": "^2.10.2"
+    "moment": "^2.10.2",
+    "cordova-plugin-local-notifications": "^1.0.0"
```
ericcf: As I mentioned, I recommend that you remove `/plugins` from `.gitignore` and add this plugin there instead.

eschlange:
@ericcf The file count is huger after my most recent update. The only files that should concern your review are: the .gitignore file and the package.json file. Thanks for your review!

ericcf:
:hurtrealbad:

PR #46, opened 2015-09-02
Ems 101782388

eschlange:
* Added continue button to cigarette log question flow.
* Added continue button to temptation log question flow.
* Changed checkboxes to radio buttons for less confusion.
* Updated specs to check for new logic.
* Removed unused functions and associated specs.

[#101782388]
[#101782328]


src/js/controllers/TemptationLogController.js
```diff
@@ -27,9 +28,8 @@
       $location.url(Routes.HOME);
     };
 
-    this.setTemptationReason = function(reason) {
-      this.temptationLog[this.TEMPTATION_REASON] = reason;
-      this.state = this.STATE_URGE;
+    this.setState = function(nextState) {
```
ericcf: Have you considered sharing code between `CigaretteLogController` and `TemptationLogController`?

eschlange: Yes, but Mark said there was a chance of requirement change which would cause them to deviate from their current form. I kept them separate for this reason.

src/partials/cigarette_log.html
```diff
@@ -23,13 +23,16 @@
             <h4 ng-if="cigarette.getCurrentlySmoking()">Why are you smoking this cigarette?</h4>
             <h4 ng-if="!cigarette.getCurrentlySmoking()">Why did you smoking this cigarette?</h4>
             <div class="row-fluid" ng-repeat="(index, value) in cigarette.SMOKING_REASONS">
```
ericcf: Looks like you're not using `index` so you could shorten this to `ng-repeat="value in cigarette.SMOKING_REASONS"`

src/js/controllers/TemptationLogController.js
ericcf: @michaelwehrley - any thoughts regarding this? I'm inclined to say it's worth refactoring to share code in such a way that it's still flexible, especially because the change hasn't come yet. :hamburger:

michaelwehrley: Seams reasonable.  If you change the constant names from `SMOKING_REASONS` and `TEMPTATION_REASONS` to `REASONS` (maybe even make a `reasonsService`) along with `saveCigaretteLog` and `saveTemptationLog` to `saveLog` you might be able to share some code.  I think that would be a worthy refactor. :-)

michaelwehrley: A possible idea: https://gist.github.com/michaelwehrley/2846f9aa7b45fe438916#file-cigarettelogcontroller-js-L25-L27

eschlange: Let's put a pin in this. :+1:

src/partials/cigarette_log.html
eschlange: Good to know, thanks!

PR #52, opened 2015-09-04
Updates for Appium

cbrenner04:


package.json
```diff
@@ -55,7 +55,7 @@
     "eslint:src": "./node_modules/.bin/eslint src",
     "eslint:test": "./node_modules/.bin/eslint test",
     "serve": "npm run build:prepare && ./node_modules/.bin/cordova run browser",
-    "test:android_features": "npm run appium:start && sleep 5 && ./node_modules/.bin/mocha test/e2e/*Spec.js && npm run appium:stop",
+    "test:android_features": "npm run android:build && npm run appium:start && sleep 5 && rspec test/features/*.rb && npm run appium:stop",
```
ericcf: I recommend updating the `appium:stop` script to: `"ps -A | grep appium | grep -v grep | egrep -o \"^\\s*\\d+\" | xargs kill"`. For medlink I found some scenarios where your existing script wouldn't kill the process.

PR #53, opened 2015-09-07
Add social support configuration functionality.

eschlange:
* Added new controller for social support addition.
* Updated configuration service to take in social supports.
* Updated setup controller and view to include social support link.
* Inserted new files into configurations.
* Added configuration specs. (they didn't exist)
* Added new social support view.
* Added integration point in session content.
* Added modal feature to sessions controller.
* Added display logic to session page.
* Added specs for the social support controller.
* Updated config logic to include social support for completeness check.
* Added files to spec config.

[#99282228]


src/js/services/Configuration.js
```diff
@@ -102,10 +110,9 @@
 
     this.configurationComplete = function () {
       return this.getCessationDate() &&
-             this.getCessationReasons() &&
-             this.getCessationReasons().length > 0 &&
-             this.getRiskyTimes() &&
-             this.getRiskyTimes().length > 0;
+             this.getCessationReasons() && this.getCessationReasons().length > 0 &&
```
ericcf: Is it necessary to test for presence and length? It looks like `getCessationReasons()` should return a default of `[]`.

src/js/services/Configuration.js
```diff
@@ -102,10 +110,9 @@
...
+             this.getRiskyTimes() && this.getRiskyTimes().length > 0 &&
```
ericcf: Same question as above.

src/js/services/Configuration.js
```diff
@@ -102,10 +110,9 @@
...
+             this.getSocialSupport() && this.getSocialSupport().length > 0;
```
ericcf: Ditto.

src/partials/session.html
```diff
@@ -32,6 +32,13 @@
                     </a>
                 </div>
 
+                <div ng-show="item.label.indexOf('[INTEGRATION|SOCIAL]') > -1">
```
ericcf: Should this logic be moved into a controller method or a directive? I find that any extra logic in the template becomes hard to maintain.

src/partials/social_support_modal.html
```diff
@@ -0,0 +1,50 @@
...
+            ng-show="social.socialSupports.length > 0"
+            ng-click="social.cancel()"><span aria-hidden="true">&times;</span></button>
+    <h3>Your Social Support</h3>
+</div>
+
+<div class="modal-body">
+    <h4>Below are the people that you can count on for supporting your decision to quit smoking:</h4>
+
+    <div class="well modal-well">
+        <span ng-hide="social.socialSupports && social.socialSupports.length > 0">Add additional social supports below.</span>
```
ericcf: Because you're testing for non-zero length of `socialSupports` in multiple places, maybe consider moving it to a method on `social`?

src/js/services/Configuration.js
eschlange: You are correct, this is unnecessary. It was likely written when the Configuration service was less stable. I'll remove them.

src/js/services/Configuration.js
eschlange: Updating.

src/js/services/Configuration.js
eschlange: Updating.

src/partials/session.html
eschlange: I agree, I'll refactor this logic into the controller.

eschlange:
@ericcf Thanks for the review! I moved the logic that you mentioned from the views to the controller, for testability and cohesion. Please take a look at my updates when you have a few minutes.

ericcf:
:monkey:

PR #54, opened 2015-09-08
Add routing logic for ema to motivation page.

eschlange:
* Updated routing logic to persist motivation notifications through EMA.
* Added specs for new methods.
* Updated view language to be more general.

[#99282204]


michaelwehrley:
:eyeglasses: good to me! :+1:

PR #55, opened 2015-09-17
Appium updates

cbrenner04:


cbrenner04:
Spent a long time struggling with setting a different date in third example so just went with the default which is today. I think this is ok for this example but might need to revisit when we need a specific date in the future or past. I also spent a fair amount of time on what would be the next example - setting a risky_time but have thus far failed to find anything that will consistently interact with the modal.

eschlange:
@cbrenner04 This looks like a good start. When we are ready to begin testing alternate cessation date logic, let's team up to try to bypass the problems you ran into.

PR #57, opened 2015-09-22
Synchronize data with dashboard.

eschlange:
* NOTE: The first several files can be ignored, they are auto generated through NPM.
* Update models to match common names for dashboard digestion.
* Update config to include authorization process.
* Added test coverage to existing specs.
* Fix existing tests to work with new synchronization logic.
* Added new models to flatten data transfer with dashboard.
* Adding spec coverage.
* Adding configuration for new and existing files.
* Updated cache registration list for synchronization.

[#102940262]


ericcf:
Is there a reason for updating the local notification plugin as part of this PR? It seems unrelated.

src/js/services/EmaAnswer.js
```diff
@@ -0,0 +1,27 @@
+(function () {
+  'use strict';
+
+  function EmaAnswer(emaAnswerCache, uuid) {
+    this.emaId = null;
```
ericcf: Is there a reason you chose to use a property assignment rather than a plain variable for `emaId`?

src/js/services/EmaAnswer.js
```diff
@@ -0,0 +1,27 @@
...
+
+    this.saveEmaAnswers = function (emaAnswers, emaId) {
+      this.emaId = emaId;
+      angular.forEach(emaAnswers, this.saveEmaAnswer, this);
+    };
+
+    this.saveEmaAnswer = function(answer) {
```
ericcf: Why did you choose a function assigned to a property rather than a plain function?

test/unit/controllers/CessationReasonsControllerSpec.js
```diff
@@ -42,16 +39,17 @@ describe('CessationReasonsController', function () {
       localStorage.clear();
     }));
 
-    it('should save a new, valid cessation reason', function () {
+    it('updates cessation reasons through config service if valid cessation reason', function () {
       controller.cessationReasonDescription = 'This is a valid cessation reason.';
+      configurationServiceMock.getCessationReasons = function() { return ['new reason'] };
       controller.saveCessationReason();
-      expect(controller.getCessationReasons().length).to.equal(1);
+      expect(configurationServiceGetCessationReasonsSpy.calledOnce).to.equal(true);
```
ericcf: How does this work if you're reassigning the property to a new function?

test/unit/services/ConfigurationSpec.js
```diff
@@ -25,97 +45,132 @@ describe('Configuration', function() {
...
+      expect(localNotificationsScheduleSpy.calledOnce).to.equal(false);
     });
-  });
 
-  describe('#saveConfigurationField', function() {
-    it('updates, deletes the old config, and persists the new data set.', function() {
-      var configValue = 'yes please';
-      var configKey = service.RISKY_TIMES_KEY;
-      someConfiguration[configKey] = configValue;
+    /*
```
ericcf: Did you forget about this commented code?

src/js/services/EmaAnswer.js
eschlange: I used this method because of the examples that I found for the Angular forEach code that I found on the internet. Do you recommend using a different approach?

ericcf: Sorry, I meant why assign `this.saveEmaAnswer` to an anonymous function, rather than just using a named function within the scope: `function saveEmaAnswer(answer) { ...`

eschlange: I use this method in order to increase flexibility with testing. If I were to use the named functions within the scope, would I be able to dynamically inject various implementations of that function using the $provide service within the Javascript specs?

test/unit/services/ConfigurationSpec.js
eschlange: I did, removing it now.

src/js/services/EmaAnswer.js
ericcf: No makes sense. Analogous to making a method public in Ruby to enable testing.

test/unit/controllers/CessationReasonsControllerSpec.js
eschlange: It took me a second to realize what was going on here. First off, I'm missing a check for the running of the save method, which I'm adding now. The reason why getCessationReasons is triggering is due to it being called upon controller initialization, before the reassignment of the function value. I'll refactor this test to make more sense.

src/js/services/EmaAnswer.js
eschlange: This piggybacks off of the logic I'm using to implement the forEach loop. I wasn't sure how to pass the parameter via the forEach function, but I knew that I could pass in the this context, which has the emaId attribute.

eschlange:
@ericcf In response to: "Is there a reason for updating the local notification plugin as part of this PR? It seems unrelated." I didn't purposely update the local notification plugin. At some point I rebuilt the package folders for Android testing and it pulled in the latest. I can try to figure out how to back out this change if you think its detrimental.

ericcf:
No biggie, just curious.

ericcf:
:soon: :paperclip:

PR #59, opened 2015-09-23
Add integration points to labels for cessation date selection.

eschlange:
* Added two integration points for cessation date selection.
* Removed temporary cessation date selection integration point.

[#104007768]


michaelwehrley:
:sweat:

PR #60, opened 2015-09-23
Add social support integration points to content.

eschlange:
* Remove temporary integration point.
* Add social support integration points.

[#104012310]


michaelwehrley:
:pizza:

PR #64, opened 2015-09-24
Fix logic for notification creation of risky times.

eschlange:
[#103810876]


src/js/controllers/RiskyTimesController.js
```diff
@@ -80,8 +80,14 @@
...
             ];
-          notification.title = 'Why you want to be smoke-free';
-          notification.text = reason.description;
+          if(reason) {
+            notification.title = 'Why you want to be smoke-free';
+            notification.text = reason.description;
+          } else {
+            notification.title = 'SiS Reminder';
+            notification.text = 'Time to enact one of your smoke-free strategies';
+            break;
```
ericcf: Is this break necessary?

eschlange: It is not, removing now.

PR #65, opened 2015-09-25
Session content update

cbitstech:
modal updates, content updates, display logic updates


eschlange:
Much code. Such logic.

PR #67, opened 2015-09-28
run eslint before unit tests

ericcf:
* this ensures that it runs automatically locally and on Travis


ericcf:
Note: this Travis failure is expected because ESlint has warnings and errors.

PR #68, opened 2015-09-28
Session content update

eschlange:
* Cleaned up the new sub session view and controller.
* Added logic to the session question service to handle sub sessions.
* Added logic to the session controller to display an ending for sub sessions.
* Updated the session content by removing an unnecessary integration point.
* Moved cache specs to their own folder.
* Added specs for the session controller.
* Fixed issues with mocks in several controller specs.
* Added spec files for cessation reason cache and service.
* Added spec files for risky times cache and service.
* Added spec files for social support cache and service.
* Added coverage for session question service methods.
* Added spec file for cessation date service.
* Added configuration to run new spec files.

[#104150782]


src/js/services/SessionQuestion.js
```diff
@@ -2,51 +2,61 @@
...
-        if(self.rawContent) {
-          self.content = self.parseRedCap(self.rawContent);
-        }
-      });
+      $http.get(this.SESSION_CONTENT_PATH).success(this.populateContentForSession);
+    };
+
+    this.populateContentForSession = function(data) {
+      var sessionFormKey = 'session_' + self.currentSessionNumber;
+      self.rawContent = $filter('filter')(data.sessionContent,
```
ericcf: You don't need to use the `self` pattern here. `this` will work because it's not being used within a closure.

eschlange: Is this the case when the method is used within the context of a callback? I'm referencing this method as a callback for the previous setSessionContent method.

test/unit/controllers/CessationReasonsControllerSpec.js
```diff
@@ -6,24 +6,34 @@ describe('CessationReasonsController', function () {
...
   beforeEach(module(function ($provide) {
     configurationServiceMock = {
       saveCessationReasons: function () {},
-      getCessationReasons: function () { return [] }
+      getCessationReasons: function () { return [] },
+      deleteCessationReason: function () { return true }
     };
     configurationServiceSaveCessationReasonsSpy = sinon.spy(configurationServiceMock, 'saveCessationReasons');
     configurationServiceGetCessationReasonsSpy = sinon.spy(configurationServiceMock, 'getCessationReasons');
+    configurationServiceDeleteCessationReasonsSpy = sinon.spy(configurationServiceMock, 'deleteCessationReason');
```
ericcf: You don't need a separate spy object when you can directly reference the spied method from the test. e.g. you could use
```javascript
describe('#deleteCessationReason', function () {
    it('should use the configuration service to delete a reason, then get the updated array', function () {
      sinon.spy(configurationServiceMock, 'deleteCessationReason');
      controller.deleteCessationReason('12345');
      expect(configurationServiceMock.deleteCessationReason.calledOnce).to.be.true;
    });
  });
```

eschlange: Cool, I did not know that this was an option!

src/js/services/SessionQuestion.js
ericcf: I see. My mistake. Yes, you either have to use the self pattern OR you could use:
```javascript
$http.get(this.SESSION_CONTENT_PATH).success(this.populateContentForSession.bind(this));
```
above.

PR #69, opened 2015-09-28
Update test files to conform to eslint rules.

eschlange:
* Updated spec files to use eslint rules.
* Removed extranious log files in git ignore.

[#104362210]


.eslintrc
```diff
@@ -13,6 +13,7 @@
     "quotes": [2, "single"],
     "no-eval": 0,
     "no-unused-vars": 1,
-    "no-multiple-empty-lines": [2, {"max": 2}]
+    "no-multiple-empty-lines": [2, {"max": 2}],
+    "global-strict": [0]
```
ericcf: Looks like this rule is deprecated: http://eslint.org/docs/rules/global-strict

Otherwise, any reason for turning off strict mode?

ericcf:
Looks cleaner to me!

eschlange:
@ericcf that eslint strict rule removal was an accident. I updated the config to include a rule that specifically checks for it.

ericcf:
:panda_face:

PR #72, opened 2015-09-29
address lack of sliders

cbitstech:
[#104174080]


src/partials/common_ema_questions.html
```diff
@@ -2,150 +2,178 @@
...
+    <input type="range" class="form-control" min="0" max="10" step="1"
             ng-model="ema.currentAnswer"
             ng-change="ema.saveQuestionAnswer('How CONFIDENT are you that you will be able to stay quit for at least 2 WEEKS?',
                                               ema.questions.STATE_CONFIDENT_MONTHS_POST)">
-        <option ng-repeat="(index, value) in ['0 - not at all confident', '1','2','3','4','5 - somewhat confident','6','7','8','9','10 - extremely confident']"
-                value="{{ index }}">
-            {{ value }}
-        </option>
-    </select>
+    <table class="slider-response">
```
ericcf: What do you think about refactoring into a slider-response directive? It would reduce a lot of nearly identical markup, and make future maintenance/changes simpler.

eschlange: @ericcf I'm going to create a story to refactor these sliders into a directive.

PR #73, opened 2015-09-29
Add notes functionality to session slides.

eschlange:
* Added textarea for slides that have "notes" as the type.
* Notes are now saved as an answer to questions when no answers are present.
* Updated the session content to remove notes from slides with integration points.
* Added spec to ensure that notes are saving through the controller.

[#104381968]


ericcf:
:chart_with_upwards_trend:

PR #74, opened 2015-09-30
Update exercise help and kindness exercise content.

eschlange:
[#104472492]


src/js/controllers/ExerciseController.js
```diff
@@ -27,12 +27,19 @@
...
       var helpItems = this.todaysExercise.help[type];
 
-      var modalInstance = $modal.open({
+      $modal.open({
         animation: true,
         templateUrl: 'myModalContent.html',
         controller: function ($scope, $modalInstance, items, type) {
             $scope.items = items;
             $scope.type = type;
+            if(type === 'exercise') {
```
michaelwehrley: If you add more types in the future, I would recommend augmenting the "item" object that is being passed in here to have a method `title` so you don't need an if/else check.

michaelwehrley:
:eyes: good! :shipit:

PR #78, opened 2015-09-30
Move smoking status up to top of home screen.

eschlange:
[#104538654]


eschlange:
Merging this in, just moving a div to a different portion of the home page.

PR #82, opened 2015-10-01
Removed logic that routes participants to EMA pages.

eschlange:
- App no longer sends participants to morning emas upon startup.
- App no longer sends partitipants to evening emas upon resume.
- Removed routing logic to motivation page after ema completion.

[#104619238]


ericcf:
:moneybag:

PR #85, opened 2015-10-05
Add data synch, add offline mode.

eschlange:
* Added config token flow back into app.
* Added offline mode by entering 'offline' as config token.
* Removed saving of session slides with just confirmation buttons.
* Stripped new 'help' value on exercises before persisting.
* Updated settings spec by removing window mock.

[#104735420]


config
```diff
@@ -10,4 +10,9 @@
     <content src="index.html" />
     <access origin="ACCESS_ORIGIN" />
     <icon src="res/icon.png" />
+
+    <access origin="*" />
+    <allow-navigation href="*" />
+    <allow-intent href="http://*/*" />
```
ericcf: Does one of the plugins use Intents?

src/js/controllers/ConfigurationController.js
```diff
@@ -19,6 +20,11 @@
           .catch(function() {
             self.alert = 'unable to configure application';
           });
+
+      if(this.configurationToken === 'offline') {
```
ericcf: Where does `this.configurationToken` get set to "offline"?

eschlange: This occurs when a user inputs the string "offline" into the configuration field.

config
eschlange: I'm checking on this now. It may not be necessary.

config
```diff
@@ -10,4 +10,9 @@
     <content src="index.html" />
     <access origin="ACCESS_ORIGIN" />
     <icon src="res/icon.png" />
+
+    <access origin="*" />
```
ericcf: If it still works with *.northwestern.edu, I'd like to see that. Otherwise seems ok.

eschlange: I removed the navigation tag, it was unnecessary. I was unable to get either the *.northwestern.edu origin or the '*://*.northwestern.edu/*' to work.

ericcf:
:leaves:

PR #87, opened 2015-10-07
Add missing character to session content.

eschlange:
[#105109476]


eschlange:
Merging for simplicity, merging for the good of all.

PR #88, opened 2015-10-08
Fix bug involving updating cessation date during session.

eschlange:
[#105110914]


michaelwehrley:
Looks good to me.  Are you planning on addressing this with any tests?

eschlange:
@michaelwehrley No, only because this logic will likely change when we add sub-sessions. This is a temporary patch until those requirements are documented.

michaelwehrley:
:+1:

PR #90, opened 2015-10-09
Update client to use cessation date object.

eschlange:
* cessation date object is now persisted to dashboard.
* replaced a simple date object with a JSON object.
* Updated specs to expect date to be nested in cessation date object.
* Updated views to expect date to be nested in cessation date object.
* Changed some default values from undefined to null.

[#105112032]


src/js/services/Session.js
```diff
@@ -13,7 +13,7 @@
     this.CESSATION_SESSION_POST_DAY_OFFSET_PLURAL = -5;
     this.CESSATION_SESSION_POST_DAY_OFFSET_SINGULAR = -6;
 
-    this.cessationDate = undefined;
+    this.cessationDate = null;
 
     this.getCessationDate = function () {
       this.cessationDate = configurationService.getCessationDate();
```
michaelwehrley: This all looks good.  In the future, would it make sense to instead call `.getCessationDate()` instead of calling the property.  For instance:
```
var cessationDay = moment(this.getCessationDate().cessationDay);
```
would become
```
var cessationDay = moment(this.getCessationDate().getCessationDay());
```
That way you are programming to the interface not the implementation.  Something to think about later.

michaelwehrley:
:+1: :eyeglasses: good!

PR #91, opened 2015-10-09
Update home page content.

eschlange:
* Removed conditional logic for status button text.
* Replaced happiness exercise button text with exercise name.

[#105326770]


eschlange:
Merging, this is a small text change for a view.

PR #93, opened 2015-10-13
Update session page to show newly structured cessation date.

eschlange:
[#105110914]


eschlange:
Small view change. Merging.

PR #94, opened 2015-10-15
Remove padding and reduce text size of exercise button.

eschlange:
[#105671444]


eschlange:
Merging style change.

PR #95, opened 2015-10-15
Fixed cig log variable names.

eschlange:
* Updated cig log variable names to match expected names.
* Removed setting cache synch, it was not being used on the server.

[#105667056]


michaelwehrley:
:+1: Whoaaa!

PR #98, opened 2015-11-30
Remove configuration token input step.

eschlange:
[#107931404]


eschlange:
Rubber-stamping to commence staging tests.

PR #100, opened 2015-12-10
Updated risky times and social support modals.

eschlange:
* Disable save buttons when no data exists.
* Added exit buttons.

[#104111644] [#104111608]


eschlange:
Merging, change consists of very small HTML updates.

PR #104, opened 2015-12-11
Update branching logic for session 2

cbrenner04:


cbrenner04:
@eschlange please review carefully in case I was a dummy

eschlange:
@cbrenner04 Looks good. We need to sit down sometime soon and document the lessons that we learned from this process so that we can build a simple solution for this problem set (branching session questions).

PR #105, opened 2015-12-15
Overhaul EMA questions and merge with cigarette log.

eschlange:
* EMA questions now match content found in associated story attachment.
* Cigarette logs now use the EMA flow and add to the common ema questions.
* References to evening and morning EMA was removed.
* Specs were modified to accomidate new logic.
* Styles and templates updated for the multi-slider ema questions.

[#104174056]


src/js/controllers/EmaController.js
```diff
@@ -7,31 +7,93 @@
...
+      motivation: 5,
+      like: 5,
+      dislike: 5
+    };
+
     this.setState = function(state) {
       this.currentQuestion = state;
     };
 
+    this.saveSliderResponses = function() {
```
michaelwehrley: Looks great; in the future - i.e., other projects -would logic like this be better put in a service?  I am not asking you to change this now though - purely a question :smiling_imp:

michaelwehrley:
:+1: merge when ready

src/js/controllers/EmaController.js
eschlange: Good question. I added it to the controller since the question data/content is stored in the view. I wanted to keep this logic out of the service because I didn't want the service knowing details about the question content. It was more a choice of keeping the question content cohesive to the front end, rather than letting it leak into the service or data layer.

This being said, if I were to go back and refactor this entire EMA question process, I'd probably store all of the questions, question meta data, and branching logic in either local storage or a database. That solution would remove the need to handle this grouped sliders in this way. Had I known the question content would undergo such significant changes, I would have architected this differently. If/when we have another project that requires a questionnaire or session questions, I'd like to take an approach where:
- The question label, answer and type are stored in a database
- The branching logic is stored in the database and associated with the questions
- There is a means for an admin to go in and update the question details and branching logic in an easy-to-use interface.

In conjunction, all of these features would remove the necessity of a programmer to update the application for any questionnaire content changes.

PR #107, opened 2015-12-16
EMA: Combine intro headers into slider pages.

eschlange:
* Moved the intro page headers onto multi slider pages.
* Updated styles for checkboxes.
* Fixed a scrolling issue with multi slider pages.

[#110262300]


src/js/controllers/EmaController.js
```diff
@@ -44,6 +44,7 @@
 
     this.setState = function(state) {
       this.currentQuestion = state;
+      $anchorScroll();
```
ericcf: Was not familiar with this!

ericcf:
:moneybag:

src/js/controllers/EmaController.js
eschlange: Took me a bit of searching to find out that it could be used to scroll to the top of the page without specifying a parameter. It can also be used to scroll to any element by specifying the element's ID as a parameter to the $anchorScroll() method.

PR #108, opened 2015-12-17
Remove day of week from risky times and add strategy.

eschlange:
* Removed the day of week from risky time creation and made the time a datetime selection.
* Updated risky time controller and edit controller to handle new data.
* Removed day specific logic.
* Refactored notification creation logic.
* Updated controller specs to handle new datetime data type.

[#110262158]


michaelwehrley:
:+1: less code looks better to me :small_red_triangle:

PR #109, opened 2015-12-20
Update branching logic in session 2 and 3

cbrenner04:


ericcf:
@cbrenner04 can you indicate what story/stories these commits are associated with? (Unless I'm being obtuse and they're already there.) Thanks!

cbrenner04:
@ericcf my bad. I always forget to do that. If it is sufficient for me to just add them here that's what I've done and I will put commit links in the stories. If not, please advise on how I can fix my mistake. Thanks

https://www.pivotaltracker.com/story/show/110446628
https://www.pivotaltracker.com/story/show/110419294
https://www.pivotaltracker.com/story/show/110301904
https://www.pivotaltracker.com/story/show/110302074

PR #114, opened 2015-12-21
Update integration disabling and session integration point.

eschlange:
* Updated session content to use the correct integration point syntax.
* Updated disable method for label integration points.
* Added a test in the controller spec to ensure new disable logic for the next button.

[#110302074]


src/js/controllers/SessionsController.js
```diff
@@ -170,10 +170,9 @@
     };
 
     this.disableNext = function(choices, label) {
-      return (choices.length > 0 &&
-             !this.currentAnswer) ||
-             (label.indexOf('[INTEGRATION') > -1 &&
-             this.configurationLock);
+      return (choices.length > 0 && !this.currentAnswer) ||
```
michaelwehrley: it took me a few times, but I figured out the logic. :smile:

michaelwehrley:
:smoking:

PR #115, opened 2015-12-22
Update EMA logic to branch on consumed substance answer.

eschlange:
[#104174056]


eschlange:
Rubber stamping for tonight's testing.

PR #117, opened 2015-12-23
Fix issue with end session integration point.

eschlange:
* Updated controller to pull raw label.
* Session question now provides the unprocessed label.
* Updated necessary specs to use new logic.
* Updated view to use new method signarture.

[#110545792]


michaelwehrley:
:+1: looks good to me.  Nicely done!

PR #119, opened 2016-01-05
Remove EMA alert if EMA is complete for current period.

eschlange:
* Updated view to use new controller logic.
* Updated ema service to pull latest complete ema.
* Updated controller logic which determines if the ema alert is visible.
* Added specs for existing and new methods in home controller and ema service.

[#110667116]


src/js/controllers/HomeController.js
```diff
@@ -52,16 +54,32 @@
...
         var notificationTime = moment(this.lastEmaNotification.at);
         var currentTime = moment();
         timeRemainingInMinutes = notificationTime.diff(currentTime, 'minutes') +
-          TIME_LIMIT_IN_MINUTES;
+          this.TIME_LIMIT_IN_MINUTES;
       }
       return timeRemainingInMinutes;
     };
+
+    this.showEmaAlert = function() {
```
michaelwehrley: I got the logic; however, in the future it maybe easier to understand what is going on here if you did something like this:
```javascript
if (noEmas()) {
  isVisible = true;
} else {
  if (didntAnswerLastEma()) {
    isVisible = true;
  }
}
```

src/js/services/Ema.js
```diff
@@ -18,6 +18,15 @@
     this.saveCigaretteLog = function(emaQuestions) {
       this.saveEma(emaQuestions, CIGARETTE_LOG_TYPE);
     };
+
+    this.getLastEma = function() {
+      var emaData = emaCache.fetchAllRaw();
+      var lastEma;
```
michaelwehrley: Did you want to set this to `null` since it may never get defined?

michaelwehrley:
:+1: looks great!

michaelwehrley:
:merge when ready! :ship:

src/js/services/Ema.js
eschlange: Good idea, less ambiguity. I'll make the update to the service and the method's spec.

src/js/controllers/HomeController.js
eschlange: Understood, thanks!

PR #121, opened 2016-01-07
Remove zero prefixes from answer.

eschlange:
[#111225280]


eschlange:
More textual changes to session content.

PR #122, opened 2016-01-07
Remove uppercase formatting of session answers.

eschlange:
[#111225498]


eschlange:
Style change to session answers.

PR #123, opened 2016-01-07
Change color of session answers.

eschlange:
[#111225846]


eschlange:
CSS color change. Merging.

PR #125, opened 2016-01-08
Add margin to session continue button.

eschlange:
[#111225910]


eschlange:
Simple style change, merging.

PR #131, opened 2016-01-11
Add ability to schedule weekly notifications for risky times.

eschlange:
* Updated view to include weekly repeat checkbox option.
* Updated controller logic to save a note to the description.
* Updated controller logic to modify use of localNotification plugin.
* Updated specs to check for updated description and clearing of option.

[#111226624]


src/js/controllers/RiskyTimesController.js
```diff
@@ -24,6 +24,7 @@
       'habitual smoking',
       'wanting something to do with your hands'
     ];
+    this.repeatWeekly = false;
```
michaelwehrley: Probably overkill here, but in the future, maybe create a `setRepeatWeekly` and a `getRepeatWeekly` function that allows you to abstract a bit more away here and use it in places like line 38:
```
if (this.getRepeatWeekly()) {...
```

src/js/controllers/RiskyTimesController.js
```diff
@@ -34,6 +35,10 @@
         localNotifications.schedule(this.generateNotification(this.riskyTime));
       }
 
+      if(this.repeatWeekly) {
+        this.riskyTime.description = this.riskyTime.description + ' (repeats weekly)';
```
michaelwehrley: Similarly, in the future use a `getRiskyTimeDescription` here - it just seems slightly strange that `this.riskyTime.description` can be an augmented value.

michaelwehrley:
:+1: looks :good!

PR #135, opened 2016-01-12
Add sub-options for each risky time strategy.

eschlange:
[#111226806]


eschlange:
Merging, small HTML content change.

PR #139, opened 2016-01-13
Added margins to different element to resolve style issues.

eschlange:
[#111465720]


eschlange:
Merging, small style change.

PR #140, opened 2016-01-20
Increase session option text size, reduced line height.

eschlange:
[#111981279]


eschlange:
Small style change, merging.

PR #141, opened 2016-01-20
Update session content for welcome screen.

eschlange:
[#111981343]


eschlange:
Merging, small session content change.

PR #142, opened 2016-01-20
Add new risky time options to edit page.

eschlange:
[#111981385]


eschlange:
Updating page content. Merging.

PR #143, opened 2016-01-20
Add consent slide to session 1 content.

eschlange:
[#111981779]


eschlange:

PR #144, opened 2016-01-20
Now forces navigation to home screen after session 1.

eschlange:
[#111982185]


eschlange:
Small page content change, merging.

PR #145, opened 2016-01-20
Add initial session 1 question to gather study ID.

eschlange:
* Added view logic to set session question.
* Added session content for additional question.

[#111982065]


eschlange:

PR #149, opened 2016-01-26
Modify notification scheduling based on session answer.

eschlange:
* Add new service and cache to store session answer data.
* Added specs for new service and cache.
* Fixed several existing specs to check for new logic.
* Fixed several formatting issues with existing specs.
* Added new file references to config files.
* Updated session controller to use new storage service.

[#111982123]


src/js/caches/LocalSessionCache.js
```diff
@@ -0,0 +1,18 @@
+(function() {
+  'use strict';
+
+  function LocalSession(cache) {
+    this.KEY = 'localSession';
```
michaelwehrley: I don't think you need this...or do you?

src/js/controllers/SessionsController.js
```diff
@@ -17,6 +17,8 @@
     this.currentSessionComplete = null;
     this.currentSessionNumber = null;
 
+    this.NUMBER_OF_CIGARETTES_QUESTION = '<h4>How many cigarettes do you smoke per day?</h4>';
```
michaelwehrley: picky, but do you need the html here or can it be in the view?

eschlange: Yes, the clinician wants the HTML tags in the questions that are saved.

src/js/caches/LocalSessionCache.js
eschlange: Yes, it is required for the inherited resource cache features.

michaelwehrley:
:all: :good: :guacamole:

eschlange:
Thanks @michaelwehrley !

PR #150, opened 2016-01-26
Remove unnecessary text from session question.

eschlange:
[#112437275]


eschlange:
Small session content change, merging.

PR #151, opened 2016-01-26
Fix syntax error in session question content.

eschlange:
[#112439891]


eschlange:
Small wording change in session question. Merging.

PR #152, opened 2016-01-26
Remove duplicate session question content.

eschlange:
[#112438543]


eschlange:
Small content change, merging.

PR #153, opened 2016-01-26
Update session content.

eschlange:
[#112438403]


eschlange:
Very small session question content change. Merging.

PR #154, opened 2016-01-26
Update session question language.

eschlange:
[#112438273]


eschlange:
Small session question content change. Merging.

PR #155, opened 2016-01-26
Reverse historical exercise view.

eschlange:
* Now shows most recent exercise results first.

[#111982235]


eschlange:
Very small change to re-order a list. Merging.

PR #156, opened 2016-01-27
Add option to ema question.

eschlange:
[#112519619]


eschlange:
Merging, small EMA content change.

PR #159, opened 2016-01-27
Remove errant characters from exercise descriptions.

eschlange:
[#112518847]


eschlange:
Small content change, merging.

PR #160, opened 2016-01-27
Remove developer note from session question.

eschlange:
[#112518113]


eschlange:
Merging, small session content change.

PR #161, opened 2016-01-27
Revise confusing session answer.

eschlange:
[#112517565]


eschlange:
Content change, merging.

PR #162, opened 2016-01-27
Revise session answer language.

eschlange:
[#112517465]


eschlange:

PR #164, opened 2016-01-27
Fix typo in session choice.

eschlange:
[#112516425]


eschlange:
Merging, small type fix.

PR #165, opened 2016-01-27
Remove dev note in session content.

eschlange:
[#112515551]


eschlange:
Small content removal, merging.

PR #167, opened 2016-01-27
Revise study ID session question.

eschlange:
[#112510977]


eschlange:
Merging, small content revision.

PR #170, opened 2016-01-28
Update number of cigarettes question specification.

eschlange:
[#112090471]


eschlange:
Very small change, merging.

PR #171, opened 2016-01-29
Add backwards navigation in sessions.

eschlange:
* Added listener to phone back button during sessions.
* Updated session controller logic for next and back buttons.
* Added backwards navigation to the session question class.
* Updated configuration logic to refresh ema notifications upon answer.
* Fixed issue with localnotification calling the wrong method when deleting notifications.
* Added and updated specs related to new functionality.
* Fixed the jquery path in the test runner.
* Updated view to now specify the question when moving onto the next question.

[#112673937]


src/js/services/Event.js
```diff
@@ -16,8 +19,14 @@
         endPoint !== EMA_MORNING_PATH &&
         endPoint !== EMA_EVENING_PATH) {
         $window.history.back();
+      } else if (endPoint === SESSION_PATH && self.sessionAction) {
+        self.sessionAction();
```
michaelwehrley: Is this what you want to run here?  i couldn't find where `sessionAction()` was defined.  Just double checking.  It seems like a property on this, not a function.

michaelwehrley:
:roller_coaster: :+1: some nice tests.

PR #172, opened 2016-02-02
Added config path to list where back button is disabled.

eschlange:
[#112673937]


eschlange:
Merging, small conditional change.

PR #175, opened 2016-05-09
update smoking cessation tools to include modal pop overs

cbitstech:
[#119226733]


src/partials/session_sub_nav.html
```diff
@@ -1,13 +1,28 @@
...
+        </a>
+      </div>
     </div>
-</div>
\ No newline at end of file
+  </div>
+</div>
+
+<script type="text/javascript">
+  $("#cessation_date_input").mobiscroll().date();
```
eschlange: Does this always set the #cessation_date_input value to the current date? Should this dynamically pull in an existing date if it exists?

eschlange:
@cbitstech aside from my one question, everything looks good. Please merge if the date loading logic is as intended.

src/partials/session_sub_nav.html
cbitstech: It should, you are absolutely right, cancelling merge.

src/js/controllers/SessionSubNavController.js
```diff
@@ -1,50 +1,56 @@
...
+    this.getCessationDate = function () {
+      return configurationService.getCessationDate();
+    };
+
+    this.setCessationDate = function () {
+      configurationService.saveCessationDate(this.cessationDate);
+    };
+
+    this.openRiskyTimesModal = function () {
+      configurationModalService.openRiskyTimesModal($modal);
```
ericcf: Is there a reason this can't be accomplished with a directive?

cbitstech: What specifically?

ericcf: Manipulating the DOM. Ideally, Angular directives should be used for that: https://github.com/johnpapa/angular-styleguide/blob/master/a1/README.md#manipulate-dom-in-a-directive

cbitstech: This is code that has been refactored/repurposed from the setup controller--I don't have the time resources to refactor it into a directive, but could definitely see the value moving forwards.

ericcf:
What's going on with this PR? Does it still need review, can it be closed, or will it be updated further? Thanks.

cbitstech:
I don't think it makes sense to turn this into a directive given the resources on this project.

cbitstech:
I'm taking this as my Saturday challenge to do this in less than 1 hour. Wish me luck!

cbitstech:
I'm issuing a new pull request and deleting this one.

PR #176, opened 2016-05-09
remove ready to begin screen

cbitstech:
[#116792999]


eschlange:
@cbitstech  👍 Nice code cleanup, please merge at your convenience.

test/unit/controllers/SessionsControllerSpec.js
```diff
@@ -169,15 +169,6 @@ describe('SessionsController', function () {
   });
 
   describe('#getSessionStarted', function () {
-    it('should return null if sessionData is not initialized.', function () {
```
ericcf: Was there a reason you added logic to this method but deleted some of its tests?

cbitstech: The biggest one is that I ultimately removed logic from the app by removing items from the view--that said, it doesn't work, so I'm cancelling it.

PR #177, opened 2016-05-14
Mjb 119226733 move setup to tools

cbitstech:
replaced all logic contained in controller with directives


src/js/directives/EnlistingSocialSupportDirective.js
```diff
@@ -0,0 +1,23 @@
...
+  function enlistingSocialSupport($modal,configurationModalService) {
+    return {
+        restrict: 'E',
+        template: 
+          '<a class="btn btn-info wide" ng-click="openSocialSupportModal()">' +
+          'Enlisting Your Social Support' +
+          '</a>',
+        link: function (scope) { 
+          scope.openSocialSupportModal = function() {
+            configurationModalService.openSocialSupportModal($modal);
```
ericcf: Is there a reason for leaving this method in a service? Would it make more sense to define it here in the directive?

src/js/directives/ScheduleQuitDayDirective.js
```diff
@@ -0,0 +1,37 @@
...
+          '<a id="cessation_date_selector" class="btn btn-info wide" onclick="$(\'#cessation_date_input\').click()">' +
+          'Schedule Your Quit Day' + 
+          '<strong> {{ getCessationDate().cessationDay | date:"MM/dd/yyyy" }} </strong>'+
+          '</a>' +
+          '<input id="cessation_date_input" type="text"'+
+          'ng-model="cessationDate.cessationDay"' +
+          'ng-change="setCessationDate()"'+
+          'mobiscroll-date />' +
+          '<script type="text/javascript">'+
+          '$("#cessation_date_input").mobiscroll().date();'+
```
ericcf: Would this work if it were moved to the `link` function? Also, is there a reason not to use Angular's built in reference to jQuery (e.g. `angular.element("#cessation_date_input").mobiscroll().date();`)?

src/js/directives/SessionSubNavDirective.js
```diff
@@ -0,0 +1,26 @@
+(function() {
+  'use strict';
+
+  function sessionNav() {
+    return {
+        restrict: 'E',
+        template: 
+        '<a ng-repeat="section in sections" ' + 
+        'class="btn btn-info wide" ' +
+        'href="#/session?startLabel={{section.startLabel}}' +
```
ericcf: There's probably a cleaner way of constructing this URL via `$location.search()`.

src/js/directives/EnlistingSocialSupportDirective.js
cbitstech: For the moment I'm leaving it because there are existing dependencies to this service from the Sessions Controller.

src/js/directives/ScheduleQuitDayDirective.js
cbitstech: Changing!

PR #178, opened 2016-05-14
Mjb 116792999 remove session beginning screen

cbitstech:
[#116792999]


src/js/controllers/SessionSubNavController.js
```diff
@@ -1,58 +1,45 @@
...
-          sessionAvailable: 1
+          sessionAvailable: 1,
+          endLabel:'Thanks do you want to use the tool again?'
         }];
 
     this.currentSessionAvailable = function() {
       return 1;
     };
+
+    $scope.sections = this.SECTIONS;
```
ericcf: Why is the `$scope` being used here? I.e. why is it necessary to mix a "controller as" pattern with scope property setting?

src/js/controllers/SessionsController.js
```diff
@@ -15,11 +15,19 @@
     this.configurationLock = true;
     this.startPosition = $routeParams.start;
     this.endPosition = $routeParams.end;
+    this.showStartScreen = $routeParams.showStart;
     this.currentSessionComplete = null;
     this.currentSessionNumber = null;
 
     this.NUMBER_OF_CIGARETTES_QUESTION = '<h4>On days that you smoke, how many cigarettes do you smoke per day?</h4>';
 
+    this.getContentAndStartSession = function(){
```
ericcf: Is there a reason this method doesn't have a unit test?

PR #179, opened 2016-05-14
Mjb 119247093 return to last page

cbitstech:
[#119247093]


src/js/controllers/SessionsController.js
```diff
@@ -116,7 +126,17 @@
       });
     };
 
+    this.skipOrShowConclusion = function(){
```
ericcf: Is there a reason this method doesn't have a unit test?

PR #180, opened 2016-05-14
Mjb 119584003 query if defaults have not changed

cbitstech:
[#119584003]


src/js/controllers/EmaController.js
```diff
@@ -26,23 +27,53 @@
...
-      dislike: 5
+      dislike: 5,
+      changedFromDefault: false
     };
 
+    this.updateChangeFromDefault = function(model){
+      model.changedFromDefault = true;
+    }
+
+    this.openEmaConfirmationModal = function () {
```
ericcf: Is there a reason not to put this logic in a directive?

src/js/controllers/EmaController.js
```diff
@@ -26,23 +27,53 @@
...
+        templateUrl: 'partials/ema_confirmation_modal.html',
+        controller: function($scope, $modalInstance){
+          $scope.close = function() {
+            $modalInstance.close();
+          };
+        }
+      })
+    };
+
+    this.proceed = function(state,model){
```
ericcf: Is there a reason not to write a unit test for this method?

cbitstech: adding!

src/js/controllers/EmaController.js
cbitstech: for the moment it is to retain consistence with the existing codebase--but agreed moving forward

